PR #17, opened 2020-06-18
Add a Discord Service using discord.js module.

noeppi-noeppi:


nodecg-io-discord/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+	"exclude": [
+		"dashboard"
+	],
```
daniel0611, line 4: The Discord bundle doesn't have a dashboard anyway so this can be safely be removed.

nodecg-io-discord/extension/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+async function validateConfig(config: DiscordServiceConfig): Promise<Result<void>> {
+    try {
+        const client = new Client()
+        return await client.login(config.botToken).then(s => {
+            client.destroy()
+            return emptySuccess();
+        }).catch(e => {
+            return error(e.toString());
+        })
```
daniel0611, line 42: I would like to keep the codebase consistent and therefore it would be nice if you could also use the async/await pattern here instead of .then and .catch. Also the catch is already handled by the `catch` block below.

nodecg-io-discord/extension/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    } catch (err) {
+        return error(err.toString());
+    }
+}
+
+function createClient(nodecg: NodeCG): (config: DiscordServiceConfig) => Promise<Result<DiscordServiceClient>> {
+    return async (config) => {
+        const client = new Client()
+        return client.login(config.botToken).then(s => {
+            client.user?.setStatus('online')
```
daniel0611, line 52: Is manually setting the online status needed? I kinda fear that it might interfere with some edge-case bundle at some point that modifies the status for some reason and after a restart it would automatically be reset to `online`.

nodecg-io-discord/extension/index.ts
```diff
@@ -0,0 +1,66 @@
...
+                    return client;
+                }
+            });
+        });
+    }
+}
+
+function stopClient(client: DiscordServiceClient): void {
+    const rawClient = client.getRawClient();
+    rawClient.user?.setStatus('invisible');
```
daniel0611, line 64: Related to the comment above.

nodecg-io-discord/extension/index.ts
```diff
@@ -0,0 +1,66 @@
...
+        return error(err.toString());
+    }
+}
+
+function createClient(nodecg: NodeCG): (config: DiscordServiceConfig) => Promise<Result<DiscordServiceClient>> {
+    return async (config) => {
+        const client = new Client()
+        return client.login(config.botToken).then(s => {
+            client.user?.setStatus('online')
+            return success({
```
daniel0611, line 53: It would be nice to have a log here that informs the user that the connection was successful. When clicking `save` in the GUI there is currently no feedback there that's why I look into the logs and a response that the config got saved and the connection is made would be helpful.

Here the log in the twitch bundle for reference:
https://github.com/codeoverflow-org/nodecg-io/blob/413d0dfaa34d9bcbf63c4c91c9ecf3a98c00b7ab/nodecg-io-twitch/extension/index.ts#L62

Issue #11, opened 2020-06-13
Enhance README


daniel0611:
### Description
The install guide, which is the current content of the README, should be moved to a wiki (#10) and the README should give information about the project structure and so on.

### References
CO issue: https://github.com/codeoverflow-org/chatoverflow/issues/123

PR #18, opened 2020-06-19
Added a simple documetation

Tim-Tech-Dev:
Added:
- install.md
- contribute.md
- twitch-chat.md

Moved  and renamed _sample_ into _samples/twitch-chat_
Refined README as mentioned in #11.


daniel0611:
This PR can be divided into two parts.

- Documentation changes
 
  I'm fine with those in general and will comment them shortly.
- Moving to `lerna link convert` and some other build changes
  
  As a total noob of lerna: whats the advantage of `lerna link convert` over `lerna bootstrap --hoist`? `lerna link convert` is kinda a hurdle for new contributors because it needs to be executed when a new npm package is created which re-generates the `package.json` to include the new package in the dependencies. 
  On the other hand `lerna bootstrap` or `lerna bootstrap --hoist`, which would solve the same problem that `lerna link convert` solves, at least to my understanding, is mostly invisible to new contributors as they, as always, just use `npm i` and `npm run build/watch`.

  Can you also explain how the new `watch`/`restartCG` script should work? I tried it, but it didn't recompile and didn't restart NodeCG. Maybe because NodeCG is blocking and `tsc-watcher` requires the onSuccess script to be non blocking? I don't know and didn't look further into this.

It would be nice to split the build stuff into a separate if you don't mind, but you can also leave it as one PR if it's too much work.

README.md
```diff
@@ -1,70 +1,62 @@
...
-## Installation
+**A NodeCG-bundle which will implement Social Media API's in the NodeCG framework**  
+>Development status: framework mostly implemented   
+>Services implemented: twitch chat, RCON
 
-Requirements: [git](https://git-scm.com/), [node](https://nodejs.org), [npm](https://www.npmjs.com/get-npm) and a working [nodecg](https://nodecg.com) installation.
+## About nodecg-io
 
-1. Clone this repository: `git clone https://github.com/codeoverflow-org/nodecg-io.git`
+NodeCG-io is the successor of [ChatOverflow](https://github.com/codeoverflow-org/chatoverflow). The aim of this bundle is to simplify the code and expand the number of integrated services. However, the main goal is still to carry out the API integration for you and thus to save you development time.
```
daniel0611, line 15: We have decided internally to spell it "nodecg-io", all lower-case. We followed the (unwritten) convention that NodeCG bundles should use only lowercase characters. Almost every NodeCG Bundle is lowercase, but not only in e.g. npm package name, but also in the Readme and everything else.

README.md
```diff
@@ -1,70 +1,62 @@
...
-   *Note 1:* This path should point to the root of this repository, not to a bundle inside this repo.
-   
-   *Note 2:* If nodecg doesn't load nodecg-io for some reason you might want to use an absolute path here.
+> <img src="https://avatars.githubusercontent.com/daniel0611"   height="50px" title="Daniel Huber"/>   | [`@daniel0611`](https://github.com/daniel0611)
 
-5. Start nodecg.
+> <img src="https://avatars.githubusercontent.com/joblo2213"    height="50px" title="Jonas"/>          | [`@joblo2213`](https://github.com/joblo2213)
 
-## Using the sample bundle
+ ><img src="https://avatars.githubusercontent.com/sebinside"    height="50px" title="Sebastian"/>      | [`@sebinside`](https://github.com/sebinside)
```
daniel0611, line 46: Small indentation error here :wink:
```suggestion
> <img src="https://avatars.githubusercontent.com/sebinside"    height="50px" title="Sebastian"/>      | [`@sebinside`](https://github.com/sebinside)
```

README.md
```diff
@@ -1,70 +1,62 @@
...
-   Select the sample bundle and the twitch service in the left bottom menu and then select the service instance that should be used by the sample bundle (in this case the name of the previously created twitch instance).
-
-7. Check the nodecg logs
-
-   You should see an error or a success message and all twitch messages that are written in the twitch channel that is hardcoded in `sample/extension/index.ts`.
+# Repository structure
 
+```
+nodecg-io/
++--docs/                   Documentaion
```
daniel0611, line 55:
```suggestion
+--docs/                   Documentation
```

docs/contribute.md
```diff
@@ -0,0 +1,98 @@
...
+Install and build all of the dependencies using `npm`:
+
+```
+cd path/to/nodecg-io 
+npm install
+```
+
+## Build
+
+### VsCode
```
daniel0611, line 35: It would be very nice if you could also commit the generated `tasks.json` so everyone could share the same settings and no additional setup is required.
I was previously using this [extension](https://marketplace.visualstudio.com/items?itemName=morissonmaciel.typescript-auto-compiler), but this, as a native editor features, seems way better!

docs/services/twitch.md
```diff
@@ -0,0 +1,35 @@
...
+
+   *Note:* If you don't see monaco on the right, try reloading the page.
+
+6. Set the created twitch service instance to the service dependency of the sample bundle.
+
+   Select the sample bundle and the twitch service in the left bottom menu and then select the service instance that should be used by the sample bundle (in this case the name of the previously created twitch instance).
+
+7. Check the nodecg logs
+
+   You should see an error or a success message and all twitch messages that are written in the twitch channel that is hardcoded in `sample/extension/index.ts`.
```
daniel0611, line 35: This path needs to be updated:
```suggestion
   You should see an error or a success message and all twitch messages that are written in the twitch channel that is hardcoded in `samples/twitch-chat/extension/index.ts`.
```

docs/package.json
```diff
@@ -0,0 +1,9 @@
+{
```
daniel0611, line 1: This is just so that NodeCG doesn't complain that this directory has no package.json and it can't start the "bundle" right?

docs/install.md
```diff
@@ -0,0 +1,54 @@
...
+```
+cd path/to/nodecg-io 
+npm run build
+```
+
+## Add nodecg-io directory to the nodecg config:
+
+Modify the nodecg configuration in `path/to/nodecg/cfg/nodecg.json`, here is an example config:
+```
+json
```
daniel0611, line 41: Language name has to go right after the triple `:
```suggestion
```json
```

nodecg-io-core/dashboard/package.json
```diff
@@ -9,7 +9,7 @@
   "dependencies": {
     "monaco-editor": "^0.20.0",
     "nodecg": "^1.6.1",
-    "nodecg-io-core": "0.1.0",
+    "nodecg-io-core": "file:..",
```
daniel0611, line 12: Next lerna noob question: Is it correct that when we run `lerna publish` it temporarily creates a package.json which contains the correct version number instead of the relative link? That way the relative path wouldn't accidentally end up in the published package.

samples/twitch-chat/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+	"extends": ".././tsconfig.common.json"
```
daniel0611, line 2:
```suggestion
	"extends": "../../tsconfig.common.json"
```

docs/install.md
```diff
@@ -0,0 +1,54 @@
+# Installation
+
+## Prerequisites
+
+In order to download necessary tools, clone the repository, and install dependencies via `npm` you need network access.
+
+You'll need the following tools:
+
+- [Git](https://git-scm.com)
+- [Node.JS](https://nodejs.org/en/)
```
daniel0611, line 10: We could add here that the minimum node version is v12, as state in the Discord.js README: https://github.com/discordjs/discord.js#installation

docs/package.json
Tim-Tech-Dev: Yes it throws an error that the folder does not contain package.jason, then it does not have the desired parameters.

daniel0611: Ok, I think we'll leave the documentation in the repo for now and move it to a other repository at some other point.

Tim-Tech-Dev:
Documentation transferred to its own [branch](https://github.com/Tim-Tech-Dev/nodecg-io/tree/documentation).

PR #23, opened 2020-06-19
Added Websocket Integration

derNiklaas:
Added websocket integration.

This was way too much effort for such little payoff.

Node module used: 
https://github.com/websockets/ws
https://www.npmjs.com/package/ws

### DIFF
--- a/nodecg-io-ws-client/extension/index.ts
+++ b/nodecg-io-ws-client/extension/index.ts
@@ -5,8 +5,7 @@ import { emptySuccess, success, error, Result } from "nodecg-io-core/extension/u
 import * as WebSocket from "ws";
 
 interface WSClientServiceConfig {
-    host: string,
-    port: number
+    address: string
 }
 
 export interface WSClientServiceClient {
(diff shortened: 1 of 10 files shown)


nodecg-io-ws-client/package-lock.json
```diff
@@ -0,0 +1,4930 @@
+{
+	"name": "nodecg-io-ws",
```
daniel0611, line 2: This is out of sync as the package has been renamed to `nodecg-io-ws-client`

nodecg-io-ws-client/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "nodecg": {
+      "compatibleRange": "^1.1.1",
+      "bundleDependencies": {
+        "nodecg-io-core": "0.1.0"
+      }
+    },
+    "license": "MIT",
+    "devDependencies": {
+      "@types/node": "^13.13.12",
+      "@types/ws": "^7.2.5",
```
daniel0611, line 27: Typings of dependencies that are exposed should also be in the normal dependencies.
This will get relevant once we publish the packages, as only packages in the normal dependencies will be installed then and the types would be missing for the bundle developer.

Source: https://www.typescriptlang.org/v2/docs/handbook/declaration-files/publishing.html#dependencies

(Also applies to `nodecg-io-ws-server`)

nodecg-io-ws-client/extension/index.ts
```diff
@@ -0,0 +1,70 @@
...
+};
+
+async function validateConfig(config: WSClientServiceConfig): Promise<Result<void>> {
+    try{
+        const client = new WebSocket(`ws://${config.host}:${config.port}`); // Will crash nodecg if the websocket couldn't connect to a server.
+        let open = false;
+        client.on('open', () => {
+            open = true;
+        });
+        await new Promise(resolve => setTimeout(resolve, 2000)); // Ugly, but I don't really know a better way
```
daniel0611, line 42: You could orient at the connect code that [rcon-client](https://github.com/janispritzkau/rcon-client/blob/dda45041562d7c81e39680d91f81187d8b17250d/src/rcon.ts#L79-L85) uses:
```typescript
await new Promise((resolve, reject) => {
    client.once("error", reject)
    client.on("open", () => {
        socket.off("error", reject)
        resolve()
    })
})
```

There you could also add the timeout the same way you did and as an advantage it wouldn't always wait 2 seconds, even if the connection has been made in a shorter time.

derNiklaas:
Wanted to change the config yesterday. Forgot about it. i will change host and port to just use address where you have to give the whole url (incl. ws:// | wss:// )

nodecg-io-ws-client/package-lock.json
derNiklaas: Will be fixed in next commit

nodecg-io-ws-client/package.json
derNiklaas: Will be fixed in next commit

derNiklaas:
Server doesn't want to support the error method 🤷

PR #29, opened 2020-06-25
IntelliJ service

noeppi-noeppi:
Add an IntelliJ connector to nodecg-io. For this a custom IntelliJ-Plugin must be installed which can be found [here](https://github.com/noeppi-noeppi/nodecg-io-intellij).
The plugin should work on all IntelliJ based IDEs but I have only tested it on IntelliJ-Community for now. There'll be more things possible in the future.

How it works:
When IntelliJ starts, it launches an HTTP-server at port 9524 (can be configured). The nodecg-io service provides a TypeScript wrapper around that HTTP-server to access IntelliJ more easily.


nodecg-io-intellij/extension/intellij.ts
```diff
@@ -0,0 +1,978 @@
...
+            sync: sync,
+            project: null,
+        });
+    }
+
+    /**
+     * Makes a raw request to IntelliJ
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
+    async rawRequest(method: string, data: any): Promise<any> {
```
daniel0611, line 90: `data` is a json object so we can use a Record to have atleast a bit better type checking.
```suggestion
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    async rawRequest(method: string, data: Record<string, unkown>): Promise<any> {
```

nodecg-io-intellij/extension/intellij.ts
```diff
@@ -0,0 +1,978 @@
...
+        }
+    }
+
+    equals(other: IntelliJ): boolean {
+        return this.address == other.address;
+    }
+}
+
+export class Project {
+    readonly intellij: IntelliJ;
```
daniel0611, line 114: This should be made private as the user isn't expected to use it and it is just for internals. This also applies to all other classes that have a intellij instance.
```suggestion
    private readonly intellij: IntelliJ;
```

noeppi-noeppi: I made the protected as there is a class that extends another (HistoryLabel) and there might be more of them in the future.

PR #33, opened 2020-06-26
Twitter service

LarsVomMars:
I wrote a quick twitter service based on [node-twitter](https://github.com/desmondmorris/node-twitter).

Some info:
- ~~It's currently untested because I need to wait for my twitter dev account to be confirmed~~
- I decided to use twitter's [oauth1.0a](https://developer.twitter.com/en/docs/basics/authentication/oauth-1-0a) instead of [oauth2 with bearer](https://developer.twitter.com/en/docs/basics/authentication/oauth-2-0) because you are [unable to interact with tweets in any way, even send](https://developer.twitter.com/en/docs/basics/authentication/overview), with the second method. This may be at the expense of [rate limits](https://developer.twitter.com/en/docs/basics/rate-limits) though.
- Every user who wants to use this service would need a [twitter dev account](https://developer.twitter.com/en/apply-for-access) due to twitters API limits.

~ Lars


LarsVomMars:
I might add a little sample and some docs later today.

LarsVomMars:
I don't see anymore typos etc., you can merge if you want and don't got any improvement proposal :)

samples/twitter-timeline/extension/index.ts
```diff
@@ -0,0 +1,30 @@
...
+            const twitterClient = client.getRawClient();
+            const params = {
+                screen_name: "skate702", // eslint-disable-line camelcase
+                exclude_replies: true, // eslint-disable-line camelcase
+                count: 50,
+            };
+            twitterClient
+                .get("statuses/user_timeline", params)
+                .then((tweets: any[]) => tweets.forEach((tweet: any) => nodecg.log.info(`Got tweet: ${tweet.text}`)))
+                .catch((err: any) => nodecg.log.error(err));
```
daniel0611, line 26: Some types can be auto-infered here and we have to ignore the last any, as we don't have any type information.
```suggestion
                // eslint-disable-next-line @typescript-eslint/no-explicit-any
                .then((tweets: any[]) => tweets.forEach((tweet) => nodecg.log.info(`Got tweet: ${tweet.text}`)))
                .catch((err) => nodecg.log.error(err));
```

nodecg-io-twitter/extension/index.ts
```diff
@@ -0,0 +1,83 @@
...
+                    return client;
+                },
+            });
+        } catch (err) {
+            return error(err.toString());
+        }
+    };
+}
+
+function stopClient(client: TwitterServiceClient): void {
```
daniel0611, line 81: Explicitly ignore the parameter here to get rid of the ESLint warning.
```suggestion
function stopClient(_client: TwitterServiceClient): void {
```

nodecg-io-twitter/extension/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+interface TwitterServiceConfig {
+    oauthConsumerKey: string;
+    oauthConsumerSecret: string;
+    oauthToken: string;
+    oauthTokenSecret: string;
+}
+
+export interface TwitterServiceClient {
+    getRawClient(): typeof Twitter;
```
daniel0611, line 15:
```suggestion
    getRawClient(): Twitter;
```

LarsVomMars: "Twitter" is of type _function_ and not a real type due to the way the twitter module is created and exported.

samples/twitter-timeline/extension/index.ts
LarsVomMars: In theory this would be of type _object_, but _object_ typings in ts aren't that great either

LarsVomMars: https://developer.twitter.com/en/docs/tweets/timelines/api-reference/get-statuses-user_timeline

nodecg-io-twitter/extension/index.ts
LarsVomMars: https://github.com/desmondmorris/node-twitter/blob/master/lib/twitter.js#L310

nodecg-io-twitter/extension/index.ts
```diff
@@ -0,0 +1,83 @@
+import { NodeCG } from "nodecg/types/server";
+import { NodeCGIOCore } from "nodecg-io-core/extension";
+import { Service, ServiceProvider } from "nodecg-io-core/extension/types";
+import { emptySuccess, success, error, Result } from "nodecg-io-core/extension/utils/result";
+const Twitter = require("twitter"); // eslint-disable-line @typescript-eslint/no-var-requires
```
daniel0611, line 5: This should be an actual import:
```suggestion
import Twitter = require("twitter");
```
This also explains why I removed the `typeof` in https://github.com/codeoverflow-org/nodecg-io/pull/33/files#r447090680. Using an actual import results in types being loaded correctly and therefore fixes the issue about it being a function which allows of getting rid of the `typeof`. I've forgotten to add this comment, sorry.

daniel0611: Also I couldn't get VS Code autocompletion working with const and the typeof and this fixed it.

LarsVomMars: I am pretty sure I tried this one and it didn't work, but maybe i messed something else up at this time.
But yes it works now...

PR #36, opened 2020-06-27
Streamdeck service

noeppi-noeppi:
Add a streamdeck service using the [elgato-stream-deck](https://www.npmjs.com/package/elgato-stream-deck) library.


noeppi-noeppi:
I'll mark this as ready for review now so it can be merged. Originally i was trying to solve the problem that only one connection to the streamdeck is possible at at one point in time. So with this you need to stop the streamdeck software before and you can't use your other buttons. I wanted to create a plugin for the native elgato streamdeck software but that is quite difficult because as soon as I started the property-inspector the debugger no longer worked... So I might try this again some time later but for now I think it's better to have this than nothing.

daniel0611:
@noeppi-noeppi you still need to remove the documentation that has been moved to https://github.com/codeoverflow-org/nodecg-io-docs/pull/2 from this PR.

PR #40, opened 2020-06-29
Added AHK Connector

derNiklaas:
This package will connect to a [HotkeylessAHK](https://github.com/sebinside/HotkeylessAHK) Server. 

using "node-fetch" package (already used for the intelliJ connector)

### DIFF
--- a/nodecg-io-ahk/extension/index.ts
+++ b/nodecg-io-ahk/extension/index.ts
@@ -45,9 +45,10 @@ async function validateConfig(config: AHKServiceConfig): Promise<Result<void>> {
 function createClient(): (config: AHKServiceConfig) => Promise<Result<AHKServiceClient>> {
     return async (config) => {
         try {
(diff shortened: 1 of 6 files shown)


nodecg-io-ahk/extension/index.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+function createClient(): (config: AHKServiceConfig) => Promise<Result<AHKServiceClient>> {
+    return async (config) => {
+        try {
+            return success({
+                getRawClient() {
+                    return new AHK(config.host, config.port);
+                },
+            });
```
daniel0611, line 52: The instance should be cached, even if this is just a http client. Otherwise I'm fine with this PR, thanks @derNiklaas.
```suggestion
            const ahk = new AHK(config.host, config.port);
            return success({
                getRawClient() {
                    return ahk;
                },
            });
```

PR #41, opened 2020-06-30
enhancement/move error catching #20

### LINKED ISSUE #20: Move error catching from Service functions to Framework
### Description
Currently all Services have a try catch block in their `validateConfig()` and `createClient()` functions.
These should be pulled into the Framework to the functions which call these to reduce code completion and to prevent Services from being able to crash NodeCG if they don't handle exceptions or maybe forget.
Also this lowers the amount of work that is required to create a Service.

Tim-Tech-Dev:
should close #20

### DIFF
--- a/nodecg-io-core/extension/instanceManager.ts
+++ b/nodecg-io-core/extension/instanceManager.ts
@@ -172,7 +172,6 @@ export class InstanceManager {
                         `The "${inst.serviceType}" service produced an error while creating a client: ${client.errorMessage}`,
                     );
                     inst.client = undefined;
-                    return;
                 } else {
                     // Update service instance object
                     inst.client = client.result;
@@ -182,7 +181,6 @@ export class InstanceManager {
                     `The "${inst.serviceType}" service function produced an error while creating a client: ${err}`,
                 );
                 inst.client = undefined;
-                return;
             }
         }
 
--- a/nodecg-io-core/extension/instanceManager.ts
+++ b/nodecg-io-core/extension/instanceManager.ts
@@ -125,7 +125,9 @@ export class InstanceManager {
                     return error("Config invalid: " + validationRes.errorMessage);
                 }
             } catch (err) {
-                return error(inst.serviceType + "Validation function invalid!: " + err);
+                this.nodecg.log.error(
+                    `The "${inst.serviceType}" service function produced an error while config verefication: ${err}`,
+                );
             }
         }
 
--- a/nodecg-io-core/extension/instanceManager.ts
+++ b/nodecg-io-core/extension/instanceManager.ts
@@ -125,9 +125,7 @@ export class InstanceManager {
                     return error("Config invalid: " + validationRes.errorMessage);
                 }
             } catch (err) {
-                this.nodecg.log.error(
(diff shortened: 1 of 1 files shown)


nodecg-io-core/extension/instanceManager.ts
```diff
@@ -119,9 +119,15 @@ export class InstanceManager {
...
-                return error("Config invalid: " + validationRes.errorMessage);
+            try {
+                const validationRes = await service.result.validateConfig(config);
+                if (validationRes.failed) {
+                    return error("Config invalid: " + validationRes.errorMessage);
+                }
+            } catch (err) {
+                this.nodecg.log.error(
+                    `The "${inst.serviceType}" service function produced an error while config verefication: ${err}`,
+                );
```
daniel0611, line 130: What was the reasoning to just log and don't return here? When the verification functions throws an error we currently just continue, set the config and create the client using a config that hasn't passed the validation function.
Also logging on the console in addition to the GUI is fine, but then we should stop and don't use the invalid config.

nodecg-io-core/extension/instanceManager.ts
```diff
@@ -159,17 +165,23 @@ export class InstanceManager {
...
+                    inst.client = undefined;
+                } else {
+                    // Update service instance object
+                    inst.client = client.result;
+                }
+            } catch (err) {
                 this.nodecg.log.error(
-                    `The "${inst.serviceType}" service produced an error while creating a client: ${client.errorMessage}`,
+                    `The "${inst.serviceType}" service function produced an error while creating a client: ${err}`,
                 );
```
daniel0611, line 184: Same as in the `updateInstanceConfig` method.
Also if the client fails to create we should unset the client same as in the if above if the service returns an error.

Tim-Tech-Dev:
No double errors in the GUI if there is no try catch block in the service implementations.
They will be removed in #42

nodecg-io-core/extension/instanceManager.ts
```diff
@@ -159,17 +163,26 @@ export class InstanceManager {
...
+            try {
+                const client = await service.result.createClient(inst.config);
+
+                // Check if a error happened while creating the client
+                if (client.failed) {
+                    this.nodecg.log.error(
+                        `The "${inst.serviceType}" service produced an error while creating a client: ${client.errorMessage}`,
+                    );
+                    inst.client = undefined;
+                    return;
```
daniel0611, line 175: We shouldn't return here and unset the client.
In case a client fails to be created the bundles should get undefined as the user did want to change the credentials and the bundles should use old ones.
```suggestion
```

nodecg-io-core/extension/instanceManager.ts
```diff
@@ -159,17 +163,26 @@ export class InstanceManager {
...
+            } catch (err) {
                 this.nodecg.log.error(
-                    `The "${inst.serviceType}" service produced an error while creating a client: ${client.errorMessage}`,
+                    `The "${inst.serviceType}" service function produced an error while creating a client: ${err}`,
                 );
                 inst.client = undefined;
-            } else {
-                // Update service instance object
-                inst.client = client.result;
+                return;
```
daniel0611, line 185:
```suggestion
```

PR #43, opened 2020-06-30
enhancement/service-boilerplate #22

### LINKED ISSUE #22: Reduce boilerplate of service implementation
### Description
There currently exists way too much boilerplate in the client implementation. 
To do this the implementation should be moved into a class which extends a class provided by the framework. This base class could do things like getting the core using a nodecg instance and printing a log if it is not loaded.

Tim-Tech-Dev:
should close #22

### DIFF
--- a/nodecg-io-core/extension/serviceBundle.ts
+++ b/nodecg-io-core/extension/serviceBundle.ts
@@ -1,41 +1,75 @@
 import { NodeCGIOCore } from ".";
 import { NodeCG } from "nodecg/types/server";
-import { Service } from "./types";
+import { Service, ServiceProvider } from "./types";
+import { emptySuccess, Result } from "./utils/result";
 
 import * as fs from "fs";
 import * as path from "path";
 
-export class serviceBundle {
+/**
+ * Class helping to create a nodecg-io service
+ */
+export class ServiceBundle {
     private core: NodeCGIOCore | undefined;
-    private service: Service<any, any>;
+    private service: Service<unknown, unknown>;
+    private nodecg: NodeCG;
 
-    constructor(nodecg: NodeCG, service: Service<any, any>) {
-        this.service = service;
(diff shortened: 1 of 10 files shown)


nodecg-io-core/extension/serviceBundle.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    async validateConfig(config: unknown): Promise<Result<void>> {
+        this.nodecg.log.error(
+            this.service.serviceType +
+                " has not provided a validateConfig method! The service won't function without it.",
+        );
+        return emptySuccess();
+    }
+
+    createClient(nodecg: NodeCG): (config: unknown) => Promise<Result<unknown>> {
```
daniel0611, line 51: Returning a function was the solution to getting access to the nodecg instance when it was a loose function.
This is not needed here because the function can access the nodecg instance through the local variable (`this.nodecg`).
Therefore we can match the function signature with the service interface and can extend it.
Also please use [generics](https://www.typescriptlang.org/docs/handbook/generics.html#generic-classes) to ensure type safety.
```suggestion
    async createClient(config: unknown): Promise<Result<unknown>> {
```

nodecg-io-core/extension/serviceBundle.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    public register(): ServiceProvider<any> | undefined {
+        if (this.core === undefined) {
+            return undefined;
+        } else {
+            return this.core.registerService(this.service);
+        }
+    }
+
+    async validateConfig(config: unknown): Promise<Result<void>> {
```
daniel0611, line 43: Instead of having a runtime error we could make the class [abstract](https://www.typescriptlang.org/docs/handbook/classes.html#abstract-classes), don't implement the function and have a runtime guarantee of a implementation by the service bundle.

PR #44, opened 2020-07-02
Philips Hue Service

TheCrether:


nodecg-io-philipshue/extension/index.ts
```diff
@@ -0,0 +1,133 @@
+import { NodeCG } from "nodecg/types/server";
+import { NodeCGIOCore } from "nodecg-io-core/extension";
+import { Service, ServiceProvider } from "nodecg-io-core/extension/types";
+import { Result, emptySuccess, error, success } from "nodecg-io-core/extension/utils/result";
+import { v4 as ipv4 } from "is-ip";
+import { v3 } from "node-hue-api";
+import Api = require("node-hue-api/lib/api/Api");
+import { UpdateInstanceConfigMessage } from "nodecg-io-core/extension/messageManager";
```
daniel0611, line 8: This can be removed as it is a leftover from trying to write the config through the framework.

nodecg-io-philipshue/extension/index.ts
```diff
@@ -0,0 +1,133 @@
+import { NodeCG } from "nodecg/types/server";
+import { NodeCGIOCore } from "nodecg-io-core/extension";
+import { Service, ServiceProvider } from "nodecg-io-core/extension/types";
+import { Result, emptySuccess, error, success } from "nodecg-io-core/extension/utils/result";
+import { v4 as ipv4 } from "is-ip";
+import { v3 } from "node-hue-api";
+import Api = require("node-hue-api/lib/api/Api");
```
daniel0611, line 7: This import is actually used but just as a typing. We can ignore the linter warning as it is actually used. 
```suggestion
// eslint-disable-next-line @typescript-eslint/no-unused-vars-experimental
import Api = require("node-hue-api/lib/api/Api");
```

nodecg-io-philipshue/package.json
```diff
@@ -0,0 +1,35 @@
...
+        }
+    },
+    "license": "MIT",
+    "devDependencies": {
+        "@types/node": "^13.13.5",
+        "nodecg": "^1.6.1",
+        "typescript": "^3.8.3"
+    },
+    "dependencies": {
+        "@types/is-ip": "^3.0.0",
```
daniel0611, line 30: `is-ip` in only used in this bundle internally but it is not exposed to the outside world, so the typings can also be a devDependency to limit the number of packages that need to be downloaded once we publish some day.

Tim-Tech-Dev:
Hey @TheCrether thanks for the PR. Since PR #42 all services use the `serviceBundle` class from `nodecg-io-core/extension/serviceBundle.ts`. You could reduce some of the code needed. Also you could maybe move the ip/port checking into the validateConfig function/method so the user can view the error in the GUI.

PR #45, opened 2020-07-02
Actions workflow for building and updating docs

joblo2213:
Workflow that automatically updates the docs once files are pushed to master.

Needs to be tested locally using [phishy/wflow](https://github.com/phishy/wflow) or [nektos/act](https://github.com/nektos/act).

Also the private access tokens for the docs repo must be generated and added as secrets to the repo as `${{ github.token }}` only has acces to the nodecg-io repo.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,40 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Update docs
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8.3
+    
+    - uses: actions/checkout@v2
+    
+    - name: Checkout docs
+      uses: actions/checkout@v2
+      with:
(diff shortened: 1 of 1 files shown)


joblo2213:
I tried to test it but I couldn't get docker to run on my windows machine as I have Windows 10 Home with update `1906` and docker requires a Pro / Enterprise / Education or Home update `2004` (which unfortunatly isn't available for my device yet).
This is couased by Hyper-V being not supported on these operating systems and WSL2 being not available. For more information on this see [Install Docker Desktop on Windows](https://docs.docker.com/docker-for-windows/install/) and [https://docs.docker.com/docker-for-windows/install-windows-home/](https://docs.docker.com/docker-for-windows/install-windows-home/)

**If somebody has docker working and set up and is familliar with github actions I'dd be happy if you could save me a lot of work by testing this.**

If nobody is found I'll need to test if I can get this to work on a linux system.

daniel0611:
I couldn't get this working with either act or wflow. Both failed at the `setup-python` action.
From looking at the workflow file the python script *should* work, but according to the [ad-m/github-push-action](https://github.com/ad-m/github-push-action#example-workflow-file)s Readme we need to manually commit and the action just pushes.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,37 @@
+
+name: Update docs
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
daniel0611, line 10: We probably should only run this on our repos and not on forks because they don't have the secrets, same as we did it with the javadoc/wiki push action of ChatOverflow.
```suggestion
    runs-on: ubuntu-latest
    if: startsWith(github.repository, 'codeoverflow-org') # don't run this in forks
```

PR #46, opened 2020-07-03
Add xdotool service

noeppi-noeppi:
Added some convenient methods that get translated into correct commands for the underlying system so it's easier to preserve platform independence.


nodecg-io-ahk/extension/AHK.ts
```diff
@@ -1,22 +1,391 @@
 import fetch from "node-fetch";
+import { KEY_MAPPING, MOUSE_MAPPING, MOUSEWHEEL_MAPPING } from "./mapping";
+import { spawn } from "child_process";
+import { onExit } from "@rauschma/stringio";
 
-export class AHK {
+export type MouseButton = "left" | "middle" | "right" | "button4" | "button5";
+export type ScrollDirection = "up" | "down" | "left" | "right";
+
+export function getAHK(mode: "ahk" | "xdotool", host: string, port: number): AHK {
```
sebinside, line 9: wrong naming. GetAHK should return AHK not something else

sebinside:
Since this is an entirely different tool, it might make sence to create a new service instead of reusing the AHK one?

noeppi-noeppi:
I thought because AHK is only running on windows and you can do many things with it that you can do with xdotool as well I would put them together into one service to make things platform independent. I understand that this sort of polymorphism makes thing harder to maintain but in my opinion platform independence is more important. But I might as well make an own service out of it if you think that's better.

PR #47, opened 2020-07-03
StreamElements Realtime Service

sebinside:
Replaces #39 by @ExtremTechniker. 

TODOs:
- [x] Refactor the original provided code
- [x] Add handlers for all common events
- [ ] Add Replicants for common events' data
- [ ] Support sub bombs
- [ ] Add sample bundle
- [ ] Test everything (it compiles already)


derNiklaas:
In my opinion the custom onSub/onCheer/etc should all use an custom EventEmitter instead of having 4 different onEvent listeners. 
I looked into the custom EventEmitter and it seems like an easy fix. I can implement it, if this is a thing you want to do.

sebinside:
Should be probably merged first and then modified, also regarding that 100+ commit PR #73.

daniel0611:
@derNiklaas letting the client extend `EventEmitter` is a interesting idea for sure!
I'll update the pr, fix linting, use the new `ServiceBundle` class and merge what we currently have till approximately tomorrow.

daniel0611:
Done. I'll merge this now so the rest of the TODOs can be made by contributors.

PR #54, opened 2020-07-06
Stop all service clients when NodeCG is stopping

### LINKED ISSUE #21: Stop all service clients when NodeCG is stopping
### Description
Currently service clients just get killed when stopping NodeCG.
Instead all clients should be gracefully stopped when NodeCG is stopping by using the `stopClient` function of the corresponding services.

noeppi-noeppi:
Closes #21

### DIFF
--- a/nodecg-io-core/extension/index.ts
+++ b/nodecg-io-core/extension/index.ts
@@ -24,6 +24,51 @@ module.exports = (nodecg: NodeCG): NodeCGIOCore => {
 
     MessageManager.registerMessageHandlers(nodecg, instanceManager, bundleManager, persistenceManager);
 
+    function onExit(): void {
+        // Unset all service instances in all bundles
+        const bundles = bundleManager.getBundleDependencies();
+        for (const bundleName in bundles) {
+            if (bundles[bundleName] !== undefined) {
+                bundles[bundleName]?.forEach((bundleDependency) => {
+                    // Only unset a service instance if it was set previously
+                    if (bundleDependency.serviceInstance !== undefined) {
+                        bundleManager.unsetServiceDependency(bundleName, bundleDependency.serviceType);
+                    }
+                });
+            }
+        }
+
+        // Call `stopClient` for all service instances
+        const instances = instanceManager.getServiceInstances();
+        for (const key in instances) {
+            if (instances[key] !== undefined) {
+                const client = instances[key]?.client;
+                const service = serviceManager.getService(instances[key]?.serviceType as string);
+                if (!service.failed) {
+                    nodecg.log.info(`Stopping service ${service.result.serviceType}.`);
+                    try {
+                        service.result.stopClient(client);
+                    } catch (err) {
+                        nodecg.log.info(`Could not stop service ${service.result.serviceType}: ${String(err)}`);
+                    }
+                }
+            }
+        }
+    }
+
+    // Normal exit
+    process.on("exit", onExit);
+    // Ctrl + C
+    process.on("SIGINT", onExit);
+    // kill
+    process.on("SIGTERM", onExit);
+    // nodemon
+    process.on("SIGUSR1", onExit);
+    process.on("SIGUSR2", onExit);
+    // Uncaught exception
+    process.on("uncaughtException", onExit);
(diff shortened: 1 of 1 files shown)


nodecg-io-core/extension/index.ts
```diff
@@ -24,6 +24,51 @@ module.exports = (nodecg: NodeCG): NodeCGIOCore => {
 
     MessageManager.registerMessageHandlers(nodecg, instanceManager, bundleManager, persistenceManager);
 
+    function onExit(): void {
```
daniel0611, line 27: I would prefer this to be an function defined at root level and not being a nested function to help with code readability.

nodecg-io-core/extension/index.ts
```diff
@@ -24,6 +24,51 @@ module.exports = (nodecg: NodeCG): NodeCGIOCore => {
...
+    // Ctrl + C
+    process.on("SIGINT", onExit);
+    // kill
+    process.on("SIGTERM", onExit);
+    // nodemon
+    process.on("SIGUSR1", onExit);
+    process.on("SIGUSR2", onExit);
+    // Uncaught exception
+    process.on("uncaughtException", onExit);
+    process.on("unhandledRejection", onExit);
```
daniel0611, line 70: Should also be extracted into its own function to help with code readability.

nodecg-io-core/extension/index.ts
```diff
@@ -24,6 +24,51 @@ module.exports = (nodecg: NodeCG): NodeCGIOCore => {
...
+        for (const bundleName in bundles) {
+            if (bundles[bundleName] !== undefined) {
+                bundles[bundleName]?.forEach((bundleDependency) => {
+                    // Only unset a service instance if it was set previously
+                    if (bundleDependency.serviceInstance !== undefined) {
+                        bundleManager.unsetServiceDependency(bundleName, bundleDependency.serviceType);
+                    }
+                });
+            }
+        }
```
daniel0611, line 39: I think you want make sure that bundles don't use the services anymore. In this case this isn't what you really want to do. You currently unset all assignments of services and the PeristanceManager saves this. So on the next start the service assignment from a bundle to a service instance is no longer there.

Instead you should use this call: `bundleDependency.clientUpdateCallback(undefined);` which does this and also doesn't overwrite the current bundle-service assignments.

nodecg-io-core/extension/index.ts
```diff
@@ -24,6 +24,51 @@ module.exports = (nodecg: NodeCG): NodeCGIOCore => {
...
+
+        // Call `stopClient` for all service instances
+        const instances = instanceManager.getServiceInstances();
+        for (const key in instances) {
+            if (instances[key] !== undefined) {
+                const client = instances[key]?.client;
+                const service = serviceManager.getService(instances[key]?.serviceType as string);
+                if (!service.failed) {
+                    nodecg.log.info(`Stopping service ${service.result.serviceType}.`);
```
daniel0611, line 48: Maybe we should also log the instance name here so we know which one stopped correctly and which service failed, if we have multiple service instances of the same type.

nodecg-io-core/extension/index.ts
```diff
@@ -24,6 +24,51 @@ module.exports = (nodecg: NodeCG): NodeCGIOCore => {
...
+
+    // Normal exit
+    process.on("exit", onExit);
+    // Ctrl + C
+    process.on("SIGINT", onExit);
+    // kill
+    process.on("SIGTERM", onExit);
+    // nodemon
+    process.on("SIGUSR1", onExit);
+    process.on("SIGUSR2", onExit);
```
daniel0611, line 67: This broke nodemon for me, as the signal was just ignored if this handler was added.
The nodemon README says that we need to re-fire the signal and also only listen for it once, so we don't create a loop: https://github.com/remy/nodemon#controlling-shutdown-of-your-script. This fixed the issue for me.

nodecg-io-core/extension/index.ts
noeppi-noeppi: I thought that the persistenceManager would save this as well but when I tested it out it actually didn't. However I changed it to what you suggested.

PR #56, opened 2020-07-07
Fixed StreamDeck Bug

ExtremTechniker:
By starting nodecg, nodecg can't find the schema, because the file name was set incorrectly.


daniel0611:
Thanks @ExtremTechniker

PR #57, opened 2020-07-07
Added Lerna Debug Log to .gitignore

ExtremTechniker:
I don't know why lerna creates this file, but its not important for GitHub repo


daniel0611:
According to the docs this file is written when an error occurs: https://github.com/lerna/lerna#concepts
Funny that I haven't encountered this even though I also had lerna runs with errors.
Thanks!

PR #58, opened 2020-07-07
Add IRC Client Service

ExtremTechniker:
Created IRC Client Service.
Fixed README. Watch for detail information in last commit `4069021`

If you wonder why 7 commit of StreamElements are in this pull request, I don't know why they are there but in the commits `bf7e1c6` and `30d3062` I deleted all files from StreamElements.


daniel0611:
Removed StreamElements commit as I worry that it will cause problems when merging #47 and to fix confilict in the README

PR #59, opened 2020-07-07
Added more functions to twitch service and bumped version

derNiklaas:
Bumped version to 0.1.1

Added onClose, onMessage, onError and send message method.


daniel0611:
I think these changes are ment for the WS service, but you modified the twitch service instead.

derNiklaas:
ah, my bad. had a rough day :D

PR #61, opened 2020-07-08
Added more functions to ws-client service

derNiklaas:
Bumped version to 0.1.1

Added onClose, onMessage, onError and send message method.
(Fixes mistakes from #59)

### DIFF
--- a/nodecg-io-ws-client/extension/index.ts
+++ b/nodecg-io-ws-client/extension/index.ts
@@ -10,6 +10,10 @@ interface WSClientServiceConfig {
 
 export interface WSClientServiceClient {
     getRawClient(): WebSocket;
+    send(message: string): void;
+    onMessage(func: (message: WebSocket.Data) => void): void;
+    onClose(func: () => void): void;
+    onError(func: () => Error): void;
 }
 
 module.exports = (nodecg: NodeCG): ServiceProvider<WSClientServiceClient> | undefined => {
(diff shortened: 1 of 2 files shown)


sebinside:
```
MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards compatible manner, and
PATCH version when you make backwards compatible bug fixes.
```

Source: https://semver.org/

derNiklaas:
so it should be bumped to 0.2.0?

nodecg-io-ws-client/extension/index.ts
```diff
@@ -10,6 +10,10 @@ interface WSClientServiceConfig {
 
 export interface WSClientServiceClient {
     getRawClient(): WebSocket;
+    send(message: string): void;
+    onMessage(func: (message: WebSocket.Data) => void): void;
+    onClose(func: () => void): void;
+    onError(func: () => Error): void;
```
daniel0611, line 16: This is wrong as the callback gets the error from the WS and not the other way arround:
```suggestion
    onError(func: (error: Error) => void): void;
```
The function definition has it correctly: https://github.com/codeoverflow-org/nodecg-io/pull/61/files#diff-d108bf9dcfde44aecbdc8d34cbea7425R61

nodecg-io-ws-client/extension/index.ts
```diff
@@ -45,6 +49,18 @@ class WSClientService extends ServiceBundle<WSClientServiceConfig, WSClientServi
...
+            },
+            onClose(func: () => void) {
+                client.on('close', func);
+            },
+            onMessage(func: (message: WebSocket.Data) => void) {
+                client.on('message', message => func(message));
+            },
+            onError(func: (error: Error) => void) {
+                client.on('error', error => func(error))
+            }
```
daniel0611, line 63: Here we can pass the handler directly and don't need to create a new function that then calls the passed function. Just makes this code a little easier to understand.
```suggestion
            onMessage(func: (message: WebSocket.Data) => void) {
                client.on('message', func);
            },
            onError(func: (error: Error) => void) {
                client.on('error', func)
            }
```

daniel0611:
Thanks @derNiklaas

PR #62, opened 2020-07-08
Add MIDI service

noeppi-noeppi:
I did not find a better way to make the `easymidi` types available for both `nodecg-io-midi-input` and `nodecg-io-midi-output` than putting them in `@types/easymidi` in the repository root.


noeppi-noeppi:
Seems like the lint task is not able to find the required libraries to build easymidi...

sebinside:
Why did you split midi input and output?

noeppi-noeppi:
As there are two different classes for input and output, the getRawClient() method would need a return type of `easymidi.Input & easymidi.Output`. The if I had a device that is only capable of output and serves no input the creation of `easymidi.Input`'d throw an error. So it's basically because `easymidi` splits it as well.

sebinside:
Puuh... Should be discussed if this is a design problem of our interface. Also regarding the desired default of getNativeClient in future versions @daniel0611

sebinside:
I mean, in this case, its probably the right decission because of the semantics of midi. but there might be other services which are split up, too. Should we support this from the framework site or just give the destinction back to the plugin dev in the form of two servcies?

daniel0611:
@sebinside I see two options for this:
1. Overthink our `getNativeClient` and maybe drop it as some services have multiple clients and also some don't have a underlying lib and their own clients, in which case there is no native client and all should be offered as functions on the ServiceClient object.
2. Multiple services can still be in one npm package/directory but they are extracted into their own files and the `extension.ts` just loads them and registers all services at the framework. The framework doesn't care if they are in one npm package or multiple packages. Then we could still distribute them as one bundle but have many services inside one bundle.

    Spliting the services might also be good for limiting access, e.g. when a bundle needs the twitch chat should't be able to access the PubSub API and by separating this way we can only give a bundle access to a single service of a whole service bundle if it has them divided.

sebinside:
There are arguments for both options but I think forcing a 1-1 relation between service and client (with `getNativeClient`) is probably the more cleaner design.

daniel0611:
The build failure can be fixed by installing [the according lib](https://packages.ubuntu.com/search?searchon=contents&keywords=asoundlib.h&mode=exactfilename&suite=artful&arch=any) in the GitHub Action. 
I'm fine with having the type definitons in this repo till the PR against the easymidi repo, which is currently open, has been merged, but with the current setup we can't publish it and bundles don't have access to it if not from inside the project.
Instead I would prefer that you just put them in the according npm package. If bundles then pull in the midi service bundle they also have the typings.

noeppi-noeppi:
I looked at the Pull Request for `easymidi` and because my typings are more detailed, I forked `easymidi` and opened another PR with improved typings. Then I put my github fork as dependency to `nodecg-io-midi-*` as a workaround till [my PR](https://github.com/dinchak/node-easymidi/pull/21) is merged. I hope that is okay.

lerna.json
```diff
@@ -1,4 +1,4 @@
 {
-    "packages": ["nodecg-io-*", "nodecg-io-core/dashboard", "samples/*"],
+    "packages": ["nodecg-io-*", "nodecg-io-core/dashboard", "samples/*", "@types/*"],
```
daniel0611, line 2: This can now be removed as the types are now in the package.
```suggestion
    "packages": ["nodecg-io-*", "nodecg-io-core/dashboard", "samples/*"],
```

nodecg-io-midi-input/extension/index.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+class MidiService extends ServiceBundle<MidiInputServiceConfig, MidiInputServiceClient> {
+    async validateConfig(config: MidiInputServiceConfig): Promise<Result<void>> {
+        try {
+            new easymidi.Input(config.device).close();
+            return emptySuccess();
+        } catch (err) {
+            return error(String(err));
+        }
```
daniel0611, line 27: This try catch block can be safely removed. The framework also now has one around this `validateConfig`/`createClient` calls (#20/#41)

nodecg-io-midi-output/extension/index.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+class MidiService extends ServiceBundle<MidiOutputServiceConfig, MidiOutputServiceClient> {
+    async validateConfig(config: MidiOutputServiceConfig): Promise<Result<void>> {
+        try {
+            new easymidi.Output(config.device).close();
+            return emptySuccess();
+        } catch (err) {
+            return error(String(err));
+        }
```
daniel0611, line 27: Same as above, this try catch block can be safely removed.

PR #68, opened 2020-07-17
Add a common service client interface to enforce getNativeClient

### LINKED ISSUE #15: Use a interface to enforce our native client requirement
### Description
We should use a interface that all service clients need to implement. This should contain a function named `getNativeClient()` that returns the original client. This way we enforce our requirement that all services need to provide access to the underlying client and we also enforce this exact name.

If we decide to add other functions that should also be available on all clients we can also add them to this interface.

Enhancement idea by @sebinside on stream and on discord.

daniel0611:
Closes #15.

- [X] Add `ServiceClient` interface and use it in all services
- [x] Make it mandatory for clients to implement the `ServiceClient` interface
- [x] Move simplification logic to external classes and let it implement the interface (as said in https://github.com/codeoverflow-org/nodecg-io/issues/15#issuecomment-657222287)
- [x] Update docs (https://github.com/codeoverflow-org/nodecg-io-docs/pull/21)


daniel0611:
I think it still unclear to me what we define as the "native client":
- If the service is using a library, e.g. the twitch service, it is obviously the client of said library.

- If the service is using a raw websocket client it should by that logic be the websocket client and not any simplification or abstraction. As an example this would apply to the StreamElements service. Using our own client implementation as the underlying client there is no reason to have any simplifications in the service as everything is implemented in the own implementation (see onSubscriber function in the StreamElements client as an example).
  Therefore these kinds of services should return the websocket client in the `getNativeClient` function, because the server and auth is provided by the instance config. Therefore a bundle can use this to do things that aren't exposed in our simplification and the `getNativeClient` has its intented usecase.

- The big unknown to me is how we want to handle services with their own clients that are built on http requests. This applies to the AHK connector as an example. What should the native client be there? All calls are theoretically simplifications of the raw http `fetch` function (or some other http lib). Returning the raw fetch function doesn't add any benefit to the bundle that is using the service.
  I think the best to return is this case is a function that wraps `fetch` and only takes the path as an argument and not the full url. That way the base url can still be set by the instance config and a bundle can actually use this to call endpoints that aren't exposed in the client.

@sebinside do you have any comments on this?

daniel0611:
I'll just leave it as it currently is and move this discussion to a new issue so that this PR can get merged at some point.
I'm going to tackle the last open point soon.

PR #70, opened 2020-07-18
Added Slack WebAPI service

ExtremTechniker:
Added Slack WebAPI service with the official Slack package (@slack/web-api)

https://npmjs.com/package/@slack/web-api

### DIFF
--- a/nodecg-io-slack/extension/index.ts
+++ b/nodecg-io-slack/extension/index.ts
@@ -1,0 +1,50 @@
+import { NodeCG } from "nodecg/types/server";
+import { ServiceProvider } from "nodecg-io-core/extension/types";
+import { emptySuccess, success, error, Result } from "nodecg-io-core/extension/utils/result";
+import { ServiceBundle } from "nodecg-io-core/extension/serviceBundle";
+import { WebClient } from "@slack/web-api";
+
+interface SlackServiceConfig {
+    token: string;
+}
+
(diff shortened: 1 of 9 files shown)


nodecg-io-slack/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "nodecg-io-slack",
+    "version": "0.1.0",
+    "description": "Allows to connect to your personal Spotify account. This enables you to e.g. control music playback or get current song information. ",
```
daniel0611, line 4: This is still the description from spotify and needs updating,

nodecg-io-slack/extension/index.ts
```diff
@@ -0,0 +1,50 @@
...
+                },
+            });
+        } else {
+            return error(res.error || "");
+        }
+    }
+
+    stopClient(client: SlackServiceClient): void {
+        client.getRawClient(); //Only that lint has no warnings
+    }
```
daniel0611, line 49: The better way to fix linting here is to have a comment inside so it passes the emptyness check and prefixing the parameter with an underscore to tell explicitly that it isn't used:
```suggestion
    stopClient(_client: SlackServiceClient): void {
        // Not supported by the client
    }
```

ExtremTechniker:
Created Docs for Slack Service in Pull Request #15 of codeoverflow-org/nodecg-io-docs

Issue #65, opened 2020-07-12
Examine Usage Scenarios for Replicants in Services


sebinside:
### Description

Replicants are a powerfull tool of NodeCG. They allow the sharing of data between multiple bundles and between backend and frontend code without requireing much boilerplate code.

This functionallity is probably useful for many service bundles, e.g. sharing the currently played song or sharing the latest subscriber. However, it shall be examined how we can use Replicants in the most efficient way. This includes the question if some functionality can be generalized and encapsulated by the nodecg-io-core.

This issue is probably also interesting for those who already used NodeCG.

A never ending discussion of @daniel0611 and @sebinside ...

### References
More on replicants: https://nodecg.com/docs/classes/replicant
Read replicants: https://nodecg.com/docs/classes/readReplicant

PR #73, opened 2020-07-26
Added Replicans and merged master into stream elements

ExtremTechniker:
I hope someone can test it because my StreamElements Account doesn’t work with nodecg-io


daniel0611:
The problem with the current straightforward implementation it that it only has one replicant, but there may be many StreamElements service instances with different accounts and therefore different channels.

As an example a bundle wants to show a normal streamoverlay for one channel and another one wants to detect subs for another channel for some reason. With this implementation the replicant would also get updated for a sub on the other channel and the streamoverlay would show subs from that other channel that it actually shouldn't be presenting.

That's what #65 is about. We plan how each bundle could setup its own replicant for its assigned service instance with (hopefully) not much boilerplate for the bundle using the service. We aren't done with that issue though.

(For the record as this was only private till now) I've had the idea that the core should have a function which takes a nodecg instance and a name for the replicant and returns a callback function which can than be used in e.g. the `onSubscriber` function. It could then be used the following way:
```typescript
streamelementsClient.onSubscriber(createReplicant(nodecg, "onSub"))
``` 
And then the calling bundle would have access to the latest subscriber using the replicant with the name `onSub`.
We haven't really discussed this idea yet internally and therefore have no final solution for this issue yet, so if you have a idea, I would happy to hear it!

PR #77, opened 2020-07-31
Added Docs as GIT Submodule

ExtremTechniker:
For more infos about Git Submodule visit https://git-scm.com/book/en/v2/Git-Tools-Submodules


joblo2213:
There has happened a lot of discussion about the benefits and drawbacks of submodules.
Currently this is not wanted.

PR #80, opened 2020-08-02
Reduce boilerplate for bundles using services

### LINKED ISSUE #31: requireService should be passed a nodecg instance instead of the bundle name
### Description
As seen in https://github.com/codeoverflow-org/nodecg-io/pull/28#discussion_r446259871 developers it is not really clear to developers how exactly the bundle name should be entered.

To fix this developer UX problem we should just require the nodecg instance to be passed to `requireService` instead of the bundle name. The nodecg instance provides a variable to get the bundleName which can then be used.

This also has the advantage that if the bundle gets renamed these references to the name get also updated automatically.
Malicious bundles can could fake the name by providing a dummy object, but that was also possibile before and way more tempting.

### References
[NodeCG bundleName variable](https://github.com/codeoverflow-org/nodecg-io/pull/28#discussion_r446259871)

### LINKED ISSUE #53: Reduce boilerplate for bundles using services
### Description
The boilerplate for bundles can be reduced by having a function in the core which gets the core and requires a service.
Also the we should return a object which can register handlers or get the raw handler instead of having to pass all handlers into the `requireService` function.

daniel0611:
Closes #53 and closes #31

- [X] Add function to reduce bundle boilerplate as noted in #53 
- [x] Update all samples to use this easier way
- [x] Remove `ServiceProvider` (aka. the old way) because it isn't needed anymore and nobody should use this anymore
- [X] Update docs (https://github.com/codeoverflow-org/nodecg-io-docs/pull/19)

### DIFF
--- a/samples/ahk/extension/index.ts
+++ b/samples/ahk/extension/index.ts
@@ -5,7 +5,7 @@ import { requireService } from "nodecg-io-core/extension/serviceClientWrapper";
 module.exports = function (nodecg: NodeCG) {
     nodecg.log.info("Sample bundle for AHK started");
 
-    const ahk = requireService<AHKServiceClient>(nodecg, "twitch");
+    const ahk = requireService<AHKServiceClient>(nodecg, "ahk");
 
     ahk?.onAvailable((client) => {
         nodecg.log.info("AHK client has been updated, sending Hello World Command.");
--- a/samples/ahk/extension/index.ts
+++ b/samples/ahk/extension/index.ts
@@ -1,20 +1,17 @@
 import { NodeCG } from "nodecg/types/server";
-import { ServiceProvider } from "nodecg-io-core/extension/types";
 import { AHKServiceClient } from "nodecg-io-ahk/extension";
+import { requireService } from "nodecg-io-core/extension/serviceClientWrapper";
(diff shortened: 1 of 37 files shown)


samples/ahk/extension/index.ts
```diff
@@ -1,20 +1,17 @@
...
-import { ServiceProvider } from "nodecg-io-core/extension/types";
 import { AHKServiceClient } from "nodecg-io-ahk/extension";
+import { requireService } from "nodecg-io-core/extension/serviceClientWrapper";
 
 module.exports = function (nodecg: NodeCG) {
     nodecg.log.info("Sample bundle for AHK started");
 
-    // This explicit cast determines the client type in the requireService call
-    const ahkClient = (nodecg.extensions["nodecg-io-ahk"] as unknown) as ServiceProvider<AHKServiceClient> | undefined;
+    const ahk = requireService<AHKServiceClient>(nodecg, "twitch");
```
LarsVomMars, line 8: Change `twitch` to `ahk`

daniel0611: Thanks for finding this. I probably wouldn't have seen that.

PR #82, opened 2020-08-06
Log a warning if a bundle requires a non-existing service

daniel0611:
Because of @derNiklaas Kappa: https://discordapp.com/channels/577412066994946060/577495219399032864/740920808792391760


derNiklaas:
Thanks.

PR #110, opened 2020-09-14
Samples/midi output send random values

StefanSchmelz:
Written a ample for midi output that sends note on messages every second


samples/midi-output/extension/index.ts
```diff
@@ -0,0 +1,29 @@
...
+    nodecg.log.info("Sample bundle for midi-output started");
+
+    const service = requireService<MidiOutputServiceClient>(nodecg, "midi-output");
+    service?.onAvailable((client) => {
+        nodecg.log.info("Midioutput client has been updated.");
+        const midiClient = client.getNativeClient();
+
+        setInterval(() => {
+            const noteVal: number = Math.round(Math.random() * 126);
+            const velocityVal: number = Math.round(Math.random() * 126);
```
daniel0611, line 16: Is there some reason that these will only be values between 0-126 instead of 0-127?

samples/midi-output/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "nodecg": {
+        "compatibleRange": "^1.1.1",
+        "bundleDependencies": {
+            "nodecg-io-midi-output": "0.1.0"
+        }
+    },
+    "scripts": {
+        "build": "tsc",
+        "watch": "tsc -w",
+        "list": "node helperscripts/listDevices.js"
```
daniel0611, line 13:
```suggestion
        "list": "node ../midi-input/helperscripts/listDevices.js"
```
It would be nice if you could start the same script in the midi-input sample here so we don't have to maintain two versions of this script.

samples/midi-input/extension/index.ts
```diff
@@ -46,6 +46,7 @@ module.exports = function (nodecg: NodeCG) {
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     function printMessage(msg: any, event: string) {
+        /*eslint-enable @typescript-eslint/no-explicit-any */
```
daniel0611, line 49: This is a left over and can be removed since I changed the ignore to only ignore the line of the function definition (667ed9a55302f45fdc61416f039550cf57664731) because it is cleaner than having a disable followed by a enable.

daniel0611 (commented):
Seems fine, only a couple nitpicks :wink:.

Can you please re-add the `/package-lock.json` from the repository root?

samples/midi-output/extension/index.ts
StefanSchmelz: No reason, just an artifact from debugging. I can change them.

daniel0611 (approved):
Thanks @StefanSchmelz

PR #117, opened 2020-09-20
Bug/midi device name fix

StefanSchmelz:
Fixed changing breaking configuration by midi device name change between reboots.

### DIFF
--- a/nodecg-io-midi-input/extension/index.ts
+++ b/nodecg-io-midi-input/extension/index.ts
@@ -38,7 +38,7 @@ class MidiService extends ServiceBundle<MidiInputServiceConfig, MidiInputService
         let devices: Array<string> = new Array<string>();
         let deviceName: string | null = null;
         easymidi.getInputs().forEach((device) => {
-            if (device.includes(config.device)) {
+            if (device.includes(config.device) && deviceName == null) {
                 deviceName = device;
             }
         });
--- a/nodecg-io-midi-input/extension/index.ts
+++ b/nodecg-io-midi-input/extension/index.ts
@@ -43,7 +43,7 @@ class MidiService extends ServiceBundle<MidiInputServiceConfig, MidiInputService
         });
 
         this.nodecg.log.info(`Connecting to MIDI input device ${deviceName}.`);
-        if (deviceName != null) {
+        if (deviceName !== null) {
             const client = new easymidi.Input(deviceName);
             this.nodecg.log.info(`Successfully connected to MIDI input device ${deviceName}.`);
             return success({
--- a/nodecg-io-midi-input/extension/index.ts
+++ b/nodecg-io-midi-input/extension/index.ts
@@ -25,7 +25,7 @@ class MidiService extends ServiceBundle<MidiInputServiceConfig, MidiInputService
         });
 
         if (devices.length === 0) {
-            return error("no device matched the configured pattern.");
+            return error("No device matched the configured pattern.");
         }
         if (devices.length > 1) {
             return error("The configured pattern is ambiguous.");
--- a/nodecg-io-midi-input/extension/index.ts
+++ b/nodecg-io-midi-input/extension/index.ts
@@ -38,7 +38,7 @@ class MidiService extends ServiceBundle<MidiInputServiceConfig, MidiInputService
         const devices: Array<string> = new Array<string>();
         let deviceName: string | null = null;
         easymidi.getInputs().forEach((device) => {
-            if (device.includes(config.device) && deviceName == null) {
+            if (device.includes(config.device) && deviceName === null) {
                 deviceName = device;
             }
         });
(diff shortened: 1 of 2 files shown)


nodecg-io-midi-input/extension/index.ts
```diff
@@ -16,12 +16,34 @@ module.exports = (nodecg: NodeCG) => {
...
-        new easymidi.Input(config.device).close();
+        const devices: Array<string> = new Array<string>();
+
+        easymidi.getInputs().forEach((device) => {
+            if (device.includes(config.device)) {
+                devices.push(device);
+            }
+        });
+
+        if (devices.length == 0) {
```
daniel0611, line 27:
```suggestion
        if (devices.length === 0) {
```
Just usual JavaScript problems

nodecg-io-midi-input/extension/index.ts
```diff
@@ -16,12 +16,34 @@ module.exports = (nodecg: NodeCG) => {
...
+            return error("no device matched the configured pattern.");
```
daniel0611, line 28: very nit: out error messages mostly start capitalized
```suggestion
            return error("No device matched the configured pattern.");
```

nodecg-io-midi-input/extension/index.ts
```diff
@@ -16,12 +16,34 @@ module.exports = (nodecg: NodeCG) => {
...
+        if (devices.length > 1) {
+            return error("The configured pattern is ambiguous.");
+        }
         return emptySuccess();
     }
 
     async createClient(config: MidiInputServiceConfig): Promise<Result<MidiInputServiceClient>> {
-        this.nodecg.log.info(`Connecting to MIDI input device ${config.device}.`);
+        this.nodecg.log.info(`Checking device name "${config.device}"`);
+        const devices: Array<string> = new Array<string>();
```
daniel0611, line 38: This is unused and can be safely removed, probably from copying the validate function.

nodecg-io-midi-input/extension/index.ts
```diff
@@ -16,12 +16,34 @@ module.exports = (nodecg: NodeCG) => {
...
+        let deviceName: string | null = null;
+        easymidi.getInputs().forEach((device) => {
+            if (device.includes(config.device) && deviceName == null) {
```
daniel0611, line 41:
```suggestion
            if (device.includes(config.device) && deviceName === null) {
```

nodecg-io-midi-input/extension/index.ts
```diff
@@ -16,12 +16,34 @@ module.exports = (nodecg: NodeCG) => {
...
+                deviceName = device;
+            }
+        });
+
+        this.nodecg.log.info(`Connecting to MIDI input device ${deviceName}.`);
         const client = new easymidi.Input(config.device);
         this.nodecg.log.info(`Successfully connected to MIDI input device ${config.device}.`);
```
daniel0611, line 48: We should also use the found device here and not the maybe incomplete string that the user has entered.
```suggestion
        const client = new easymidi.Input(deviceName);
        this.nodecg.log.info(`Successfully connected to MIDI input device ${deviceName}.`);
```
You also need to check `deviceName` for null. The framework doesn't call validate when loading the config from disk because it assumes that a config which was once valid will be always valid. This was done to improve startup performance but I'm thinking of changing it back because it creates this hassle for service devs.

daniel0611 (changes requested):
Thanks a lot for this in advance! I've found a couple of problems as stated below. (The comments also regard the output but I didn't want to write it twice.)

nodecg-io-midi-input/extension/index.ts
StefanSchmelz: I'll fix that for now.

StefanSchmelz:
hope i fidex everything

nodecg-io-midi-input/extension/index.ts
StefanSchmelz: added a little check to enshure this is not going to break.

daniel0611:
@StefanSchmelz can you please also fix the double vs triple slashes in the midi output?

nodecg-io-midi-input/extension/index.ts
```diff
@@ -16,20 +16,44 @@ module.exports = (nodecg: NodeCG) => {
...
+        easymidi.getInputs().forEach((device) => {
+            if (device.includes(config.device) && deviceName === null) {
+                deviceName = device;
+            }
         });
+
+        this.nodecg.log.info(`Connecting to MIDI input device ${deviceName}.`);
+        if (deviceName != null) {
+            const client = new easymidi.Input(deviceName);
+            this.nodecg.log.info(`Successfully connected to MIDI input device ${deviceName}.`);
```
daniel0611, line 48: This should also be a triple equals

StefanSchmelz: You mean !=== 
??

StefanSchmelz: nvm i figured it out

daniel0611 (approved):
Thanks a lot for this!

Issue #111, opened 2020-09-15
Add missing basic sample bundles


daniel0611:
### Description
It would be nice if we could have for each service at least a very simple sample bundle.

Services with currently no sample at all:
- [ ] IRC
- [x] RCON
- [x] sACN sender
- [x] Spotify
- [x] Websocket client

PR #119, opened 2020-09-21
Created websocket sample

StefanSchmelz:
Created a sample that connects to a websocket server and sends a ping every second.

### DIFF
--- a/samples/websocket-client/extension/index.ts
+++ b/samples/websocket-client/extension/index.ts
@@ -1,0 +1,23 @@
+import { NodeCG } from "nodecg/types/server";
+import { requireService } from "nodecg-io-core/extension/serviceClientWrapper";
+import { WSClientServiceClient } from "nodecg-io-websocket-client/extension";
+
+module.exports = function (nodecg: NodeCG) {
+    nodecg.log.info("Sample bundle for <the-service-name> started");
+
+    const service = requireService<WSClientServiceClient>(nodecg, "websocket-client");
+    service?.onAvailable((client) => {
+        nodecg.log.info("Client has been updated.");
+
+        const sock = client.getNativeClient();
+        sock.addListener("message", (code: number, data: string) => {
+            nodecg.log.info(`recieved message "${data}" with code: ${code}`);
+        });
+        setInterval(() => {
+            nodecg.log.info("Sending ping ...");
+            sock.send("ping");
+        }, 1000);
+    });
+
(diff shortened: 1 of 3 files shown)


StefanSchmelz:
This should be part of the issue #111

samples/websocket-client/extension/index.ts
```diff
@@ -0,0 +1,23 @@
+import { NodeCG } from "nodecg/types/server";
+import { requireService } from "nodecg-io-core/extension/serviceClientWrapper";
+import { WSClientServiceClient } from "nodecg-io-websocket-client/extension";
+
+module.exports = function (nodecg: NodeCG) {
+    nodecg.log.info("Sample bundle for <the-service-name> started");
```
daniel0611, line 6:
```suggestion
    nodecg.log.info("Sample bundle for websocket-client started");
```

samples/websocket-client/extension/index.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    const service = requireService<WSClientServiceClient>(nodecg, "websocket-client");
+    service?.onAvailable((client) => {
+        nodecg.log.info("Client has been updated.");
+
+        const sock = client.getNativeClient();
```
daniel0611, line 12: Instead of using the native client it would be nice if you use the simplifications of the client here.
E.g. `client.send` and `client.onMessage`

samples/websocket-client/extension/index.ts
```diff
@@ -0,0 +1,23 @@
...
+        sock.addListener("message", (code: number, data: string) => {
+            nodecg.log.info(`recieved message "${data}" with code: ${code}`);
+        });
+        setInterval(() => {
+            nodecg.log.info("Sending ping ...");
+            sock.send("ping");
+        }, 1000);
```
daniel0611, line 19: I would prefer this to be more like 10 seconds so it doesn't spam the log as badly.
```suggestion
        }, 10000);
```

daniel0611 (changes requested):

samples/websocket-client/extension/index.ts
StefanSchmelz: I'll do that

StefanSchmelz: changed. resolving

StefanSchmelz:
Hope that was everything and I did not make any dumb misstakes

daniel0611 (approved):
Thanks

PR #122, opened 2020-09-22
Unregister handlers of bundle when assigning an other instance

### LINKED ISSUE #9: Unregister handlers of bundle when assigning an other instance
### Description
Currently there is no way to unregister handlers which a bundle has registered.

Given this scenario:
1. Have the sample twitch bundle loaded with no assigned service instance and a created twitch service instance.
2. Assign the instance to the sample bundle and see that the twitch chat is printed correctly.
3. Unassign the service instance from the bundle (setting it to "none" in the GUI), notice how the chat is still printed.
4. Reassign the service instance and notice how each message is printed twice.

This is because the bundle has registered its handler when assigning the service instance for its first time. When the service instance is unassigned the handler still lives because there is no way for the bundle to unregister its handler. When it is reassigned two handlers are registered causing the above mentioned behavior.

This is obviously fixed when you restart nodecg, but as a small workaround you can also click `Save` on the service instance triggering a recreation of the service client followed by a registering of the handlers by each bundle which uses this service instance which results in only one attached handler.

Here some ideas how we could fix this:

1. Create one service client for each instance that uses it. This also fixes that a different bundle could do nasty stuff to disturb other bundles like randomly stopping the service client or unregistering all handlers, but on the other side this can create problems with services like Spotify which only allow a limited amount of unique connections.
2. The bundle should supply a callback which removes the registered handlers from the given service instance. Bad because it isn't enforced and bundle developers might forget this and have to remember this when they create a new handler for some event.
3. Have the service provide a function which removes all listeners. Then if the assigned service instance to a bundle changes this function would be called which would unregister all event listeners. This client with no registered handlers would then be passed to all bundles that use this instance again so they can add their listeners again.

I'm not really happy with all of these solutions, but 1 and 3 would be acceptable I guess.

Note: services that don't make use of event callbacks wouldn't be affected. This might be services like [OLA](https://wiki.openlighting.org/index.php/OLA_JSON_API) that are just HTTP and don't have any sockets and therefore no events.

daniel0611:
Closes #9

- [X] implement `removeHandlers` in the service interface and adapt framework to use it
- [x] provide a  `removeHandlers` function for services that need it


daniel0611:
The telegram client changes are now upstreamed, but still need typing updates. Otherwise this PR is finally done.

PR #123, opened 2020-09-26
Add a way to get PID of IntelliJ

noeppi-noeppi:
I just realised that I added this feature to the intellij plugin more than two month ago and never opened a pull request for this here.


daniel0611 (approved):
Thanks!

PR #125, opened 2020-09-26
Add a tiane service

noeppi-noeppi:
TIANE is a german open smart-home-speech-assistant. https://github.com/FerdiKr/TIANE
To find out how to use this you can take a look here https://github.com/FerdiKr/TIANE/blob/master/WebSocket.md


daniel0611 (approved):
Thanks! I guess you'll add the sample documentation later.

PR #127, opened 2020-09-30
Added simple rcon sample

derNiklaas:
This is a simple rcon-minecraft sample.
It sends /list to a minecraft server, prints the output and then performs /say nodecg-io speaking here!


daniel0611 (approved):
Thanks!

PR #128, opened 2020-10-20
Feature/serial service

NeoCortex97:
I created a basic service that should be able to talk to a serial device. Sample will get its own pr.

### DIFF
--- a/nodecg-io-serial/extension/SerialClient.ts
+++ b/nodecg-io-serial/extension/SerialClient.ts
@@ -46,8 +46,7 @@ export class SerialServiceClient implements ServiceClient<SerialPort> {
         if (deviceInfo.port) {
             result.push(deviceInfo.port);
         } else {
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            devices.forEach((element: any) => {
+            devices.forEach((element) => {
                 if (deviceInfo.pnpId && "pnpId" in element && element.pnpId === deviceInfo.pnpId) {
                     result.push(element["path"]);
                 } else if (
--- a/nodecg-io-serial/extension/SerialClient.ts
+++ b/nodecg-io-serial/extension/SerialClient.ts
@@ -10,8 +10,8 @@ export interface DeviceInfo {
 }
 
 interface Protocol {
-    delimiter: string;
-    encoding: string;
+    delimiter: "\n\r" | "\n";
+    encoding: "ascii" | "utf8" | "utf16le" | "ucs2" | "base64" | "binary" | "hex" | undefined;
 }
 
 export interface SerialServiceConfig {
--- a/nodecg-io-serial/extension/SerialClient.ts
+++ b/nodecg-io-serial/extension/SerialClient.ts
@@ -64,7 +64,7 @@ export class SerialServiceClient implements ServiceClient<SerialPort> {
         if (result.length < 1) {
             return error("No device matched the provided criteria.");
         } else if (result.length > 1) {
-            return error("The provided criteria were abiguous.");
+            return error("The provided criteria were ambiguous.");
         } else {
             return success(result[0]);
         }
--- a/nodecg-io-serial/extension/SerialClient.ts
+++ b/nodecg-io-serial/extension/SerialClient.ts
@@ -86,7 +86,7 @@ export class SerialServiceClient implements ServiceClient<SerialPort> {
     }
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    onData(callback: (...args: any[]) => void): void {
+    onData(callback: (data: any) => void): void {
         this.parser.on("data", callback);
     }
 }
--- a/nodecg-io-serial/extension/SerialClient.ts
+++ b/nodecg-io-serial/extension/SerialClient.ts
@@ -1,0 +1,93 @@
+import { ServiceClient } from "nodecg-io-core/extension/types";
+import { emptySuccess, success, error, Result } from "nodecg-io-core/extension/utils/result";
+import { ReadLine } from "readline";
+import SerialPort = require("serialport"); // This is neccesary, because serialport only likes require!
+// This is neccesary because serialport does not work with import...
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const Readline = require("@serialport/parser-readline");
+
+export interface DeviceInfo {
+    port: string;
+    manucaturer: string;
+    serialNumber: string;
+    pnpId: string;
+}
+
+interface Protocoll {
+    delimiter: string;
+    encoding: string;
+}
+
+export interface SerialServiceConfig {
+    device: DeviceInfo;
+    connection: SerialPort.OpenOptions;
+    protocoll: Protocoll;
+}
+
+export class SerialServiceClient implements ServiceClient<SerialPort> {
+    private serialPort: SerialPort;
+    private parser: ReadLine;
+    constructor(settings: SerialServiceConfig) {
+        let port = "";
+        SerialServiceClient.inferPort(settings.device).then((value) => {
+            port = value;
+        });
+        this.serialPort = new SerialPort(port, settings.connection);
+        this.parser = this.serialPort.pipe(new Readline(settings.protocoll));
+    }
+
+    getNativeClient(): SerialPort {
+        return this.serialPort;
+    }
+
+    static async inferPort(deviceInfo: DeviceInfo): Promise<string> {
+        const result = [];
+        const devices = await SerialPort.list();
+        if ("port" in deviceInfo) {
+            result.push(deviceInfo.port);
+        } else {
+            // hope this ignore is right.
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            devices.forEach((element: any) => {
+                console.log("path" in element);
+                if ("pnpId" in deviceInfo && "pnpId" in element && element.pnpId === deviceInfo["pnpId"]) {
+                    console.log(element);
+                    result.push(element.path);
+                } else if (
+                    "manufacturer" in deviceInfo &&
+                    "serialNumber" in deviceInfo &&
+                    element.manufacturer === deviceInfo["manufacturer"] &&
+                    element.serialNumber === deviceInfo["serialNumber"]
+                ) {
+                    console.log(element);
+                    result.push(element.path);
+                }
+            });
+        }
+
+        // Check if result isn't empty or ambiguous
+        if (result.length < 1) {
+            return "";
+        } else if (result.length > 1) {
+            return "";
+        } else {
+            return result[0];
+        }
+    }
+
+    close(): void {
+        this.serialPort.close();
+    }
+
+    send(payload: string): void {
+        this.serialPort.write(payload, (err) => {
+            if (err) {
+                throw err;
+            }
+        });
+    }
+
+    onData(callback: (param: string) => string): void {
+        this.parser.on("data", callback);
+    }
+}
--- a/nodecg-io-serial/extension/SerialClient.ts
+++ b/nodecg-io-serial/extension/SerialClient.ts
@@ -74,14 +74,18 @@ export class SerialServiceClient implements ServiceClient<SerialPort> {
         this.serialPort.close();
     }
 
-    send(payload: string): Result<string> {
-        let res = success("OK") as Result<string>;
-        this.serialPort.write(payload, (err) => {
-            if (err) {
-                res = error(err);
-            }
+    async send(payload: string): Promise<Result<string>> {
+        const err: Error | undefined | null = await new Promise((resolve) => {
+            this.serialPort.write(payload, (err) => {
+                resolve(err);
+            });
         });
-        return res;
+
+        if (err) {
+            return error(err.message);
+        } else {
+            return success("OK");
+        }
     }
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/nodecg-io-serial/extension/SerialClient.ts
+++ b/nodecg-io-serial/extension/SerialClient.ts
@@ -1,5 +1,5 @@
 import { ServiceClient } from "nodecg-io-core/extension/types";
-import { success, error, Result, emptySuccess, Success } from "nodecg-io-core/extension/utils/result";
+import { success, error, Result, emptySuccess } from "nodecg-io-core/extension/utils/result";
 import * as SerialPort from "serialport"; // This is neccesary, because serialport only likes require!
 
 export interface DeviceInfo {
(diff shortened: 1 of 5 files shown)


nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,92 @@
...
+        this.serialPort.write(payload, (err) => {
+            if (err) {
+                res = error(err);
+            }
+        });
+        return res;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onData(callback: (...args: any[]) => void): void {
```
daniel0611, line 89: The `parser.on` method passes just one argument which contains the message so the array would always have a length of 1. Instead we should just pass one parameter with the data.
```suggestion
    onData(callback: (data: any) => void): void {
```

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,92 @@
+import { ServiceClient } from "nodecg-io-core/extension/types";
+import { success, error, Result } from "nodecg-io-core/extension/utils/result";
+import { ReadLine } from "readline";
+import SerialPort = require("serialport"); // This is neccesary, because serialport only likes require!
+// This is neccesary because serialport does not work with import...
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const Readline = require("@serialport/parser-readline");
```
daniel0611, line 7: By using `const` here instead of `import` `Readline` has no types.
Instead you should use `SerialPort.parsers.Readline` which is the way to get access to the class and also the typings.

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export interface DeviceInfo {
+    port: string;
+    manucaturer: string;
+    serialNumber: string;
+    pnpId: string;
+}
+
+interface Protocoll {
```
daniel0611, line 16: Small typing here :wink: (also applies to the rest) 
```suggestion
interface Protocol {
```

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,92 @@
...
+    delimiter: string;
+    encoding: string;
```
daniel0611, line 18: If you use `SerialPort.parsers.Readline` as I above recommended the compiler will complain because encoding has to be one off these strings instead of any string. You may also want to update this in the json schema.
```suggestion
    encoding: 'ascii' | 'utf8' | 'utf16le' | 'ucs2' | 'base64' | 'binary' | 'hex';
```

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,92 @@
...
+export class SerialServiceClient implements ServiceClient<SerialPort> {
+    private serialPort: SerialPort;
+    private parser: ReadLine;
+    constructor(settings: SerialServiceConfig) {
+        SerialServiceClient.inferPort(settings.device).then((port) => {
+            if (!port.failed) {
+                this.serialPort = new SerialPort(port.result, settings.connection);
+                this.parser = this.serialPort.pipe(new Readline(settings.protocoll));
+            }
+        });
```
daniel0611, line 36: This is problematic because we don't wait for the promise. We call the constructor and create the promise but we return instantly and the promise may throw an exception later which would completely crash nodecg (if we are waiting for it a `try` inside the framework would catch it).

To be honest I would just do this in a seperate method (e.g. `init`) that then could be `async` and `await` the result. And then we can call the `init` method it in the `createClient` function and also properly handle the error if there is any.

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    static async inferPort(deviceInfo: DeviceInfo): Promise<Result<string>> {
+        const result = [];
+        const devices = await SerialPort.list();
+        if ("port" in deviceInfo) {
+            result.push(deviceInfo.port);
+        } else {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            devices.forEach((element: any) => {
```
daniel0611, line 50:
```suggestion
            devices.forEach((element) => {
```
The element is of type `SerialPort.PortInfo` but TypeScript will infer that type automatically when we provide none.

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,92 @@
...
+                    result.push(element["path"]);
+                }
+            });
+        }
+
+        // Check if result isn't empty or ambiguous
+        if (result.length < 1) {
+            return error("No device matched the provided criteria.");
+        } else if (result.length > 1) {
+            return error("The provided criteria were abiguous.");
```
daniel0611, line 68:
```suggestion
            return error("The provided criteria were ambiguous.");
```

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    send(payload: string): Result<string> {
+        let res = success("OK") as Result<string>;
+        this.serialPort.write(payload, (err) => {
+            if (err) {
+                res = error(err);
+            }
+        });
+        return res;
+    }
```
daniel0611, line 86: The `write` function returns immediately and writes the data at some point later. So it will return now and the send function will return the "OK" message. Then once it wrote the data to the serial port the callback will get executed and the `res` variable will be set, but the return was way before that. Because of that this function will always return "OK" and doesn't work as you might have intended.

Here's the way it would solve this: we create a new promise and call the `write` function inside the promise.
Once the data has been written we call the `resolve` function which will mark the promise as finished and also give the promise the return value of the `err` variable. Then we get the value in the `err` variable of the function which will have waited for that because it has `await`ed the promise.
The `Error | undefined | null` type is required because that is also the type of the `err` variable from the `write` callback.

```suggestion
    async send(payload: string): Promise<Result<string>> {
        const err: Error | undefined | null = await new Promise((resolve) => {
            this.serialPort.write(payload, (err) => {
                resolve(err);
            });
        });

        if (err) {
            return error(err.message);
        } else {
            return success("OK");
        }
    }
```

nodecg-io-serial/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "nodecg": {
+        "compatibleRange": "^1.1.1",
+        "bundleDependencies": {
+            "nodecg-io-core": "0.1.0"
+        }
+    },
+    "license": "MIT",
+    "devDependencies": {
+        "@types/node": "^14.6.4",
+        "@types/serialport": "^6.0.1",
```
daniel0611, line 28: Is there some reason we need `6.0.1`? I would prefer to have a newer version here also because we're using 9.0.1 of the serialport lib.
```suggestion
        "@types/serialport": "^8.0.1",
```

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    getNativeClient(): SerialPort {
+        return this.serialPort;
+    }
+
+    static async inferPort(deviceInfo: DeviceInfo): Promise<Result<string>> {
+        const result = [];
+        const devices = await SerialPort.list();
+        if ("port" in deviceInfo) {
```
daniel0611, line 46: I would prefer to access the property directly (`deviceInfo.port`) because that way the compile will make sure that the name of the variable has no typo in it. This works because if the property is not defined it will be `undefined` which is a falsey value and any real value will be truthy.

daniel0611 (changes requested):
Sorry that it took me so long to review your PR :sweat_smile:.
I hope you'll understand my comments, if not feel free to ask for clarification.
Also thanks for this in advance even tho there are still some things to be done!

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: I just copied the signature form the parser.on method.

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: Thanks, will be fixed

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: Whoops, I originally only added the most common encodings like ascii, utf8, binary and hex in the schema and did not think about the interface... 
I totally agree with you.

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: I agree with you on that, but did not think about that myself. I'll implement that.

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: This was a leftover from a much earlier version...
The compiler was complaining that the compiler was unable to properly infer the type. This was my attempt to tell the compiler don't worry about that.

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: Your suggestion was pretty much what I intended.
I got thrown over once again by javascript being highly asynchronous...
I was not aware that write was async.

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: I used this method because I hit some resistance from the compiler...
The compiler was complaining that the attribute was not defined.
I'll fix this everywhere.

nodecg-io-serial/package.json
StefanSchmelz: Whoops... This happens if you copy stuff from documentation... Maybe this was giving me these weird type errors.

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: I tried that, but the compiler refused to produce any code...
Maybe this was caused by me using old typings. I'll test that

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: After accepting your change the build fails now get the following compiler error:
```
extension/SerialClient.ts(49,30): error TS7006: Parameter 'element' implicitly has an 'any' type.
```
This is why I originally put any there.

daniel0611: Your latest commit with the updated typings fixed CI so it seems that was too due to outdated typings.

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: I fixed it for all usages of deviceInfo. But I can't atm give the element a proper type, because typescript states that it does not know Portinfo...
I'll look further into that later tho

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: I was able to remove that import completely. Thanks.

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,95 @@
...
+    connection: SerialPort.OpenOptions;
+    protocol: Protocol;
+}
+
+export class SerialServiceClient implements ServiceClient<SerialPort> {
+    private serialPort: SerialPort;
+    private parser: SerialPort.parsers.Readline;
+    constructor(private readonly settings: SerialServiceConfig) {}
+
+    async init(): Promise<void> {
```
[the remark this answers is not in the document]
daniel0611: This also seems to be the reason for the promise rejections.
If you create a client which isn't used the `init` function is never called, therefore is `serialPort` undefined.
Then when the client gets closed(e.g. because you shutdown nodecg or change the config of the serial service instance) `stopClient()` of the service calls `close()` which will try to call `this.serialPort.close()` but because `serialPort` is null if `init` is never called this fails catastrophically.

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,95 @@
...
+                if (deviceInfo.pnpId && "pnpId" in element && element.pnpId === deviceInfo.pnpId) {
+                    result.push(element["path"]);
+                } else if (
+                    deviceInfo.manucaturer &&
+                    deviceInfo.serialNumber &&
+                    element.manufacturer === deviceInfo.manucaturer &&
+                    element.serialNumber === deviceInfo.serialNumber
+                ) {
+                    result.push(element["path"]);
+                }
```
daniel0611, line 59: nit: use direct accessing of properties instead of strings because they can be type checked.
```suggestion
                if (deviceInfo.pnpId && element.pnpId && element.pnpId === deviceInfo.pnpId) {
                    result.push(element.path);
                } else if (
                    deviceInfo.manucaturer &&
                    deviceInfo.serialNumber &&
                    element.manufacturer === deviceInfo.manucaturer &&
                    element.serialNumber === deviceInfo.serialNumber
                ) {
                    result.push(element.path);
                }
```

daniel0611 (changes requested):

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: I tried it, but the compiler said that it the attribute does not exist and it can not infer the type of element. Therefore it tells me that element has an implicit type of any....
I don't really know how to fix that in a clean way besides importing the types of serialport directly.

But I totally agree with you that it should be like that!

nodecg-io-serial/extension/SerialClient.ts
```diff
@@ -0,0 +1,95 @@
...
+    constructor(private readonly settings: SerialServiceConfig) {}
+
+    async init(): Promise<void> {
+        const port = await SerialServiceClient.inferPort(this.settings.device);
+        if (!port.failed) {
+            this.serialPort = new SerialPort(port.result, this.settings.connection);
+            this.parser = this.serialPort.pipe(new SerialPort.parsers.Readline(this.settings.protocol));
+        } else {
+            console.log("Could not initialize the serial port");
+            // TODO: handle error properly
```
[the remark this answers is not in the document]
StefanSchmelz: You are totally right.

nodecg-io-serial/extension/SerialClient.ts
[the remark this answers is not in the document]
StefanSchmelz: I hope i fixed this correctly

nodecg-io-serial/extension/SerialClient.ts
StefanSchmelz: nvm.... I tried it again and it works fine

StefanSchmelz:
Sry that I messed up with the branches.

daniel0611 (approved):
Looks good now! Thanks for this PR, will be merged soon

PR #132, opened 2020-11-02
Sample/serial ping pong

NeoCortex97:
Created a sample for my serial service that should be merged in pr #128


daniel0611 (approved):
The call of `init()` should be removed after my review comments of #128 are addressed but otherwise it looks good! Thanks. 
Will be merged after #128 has been merged.

PR #137, opened 2020-11-10
Add a reddit service

noeppi-noeppi:
Currently `reddit-ts` is installed from my github because of an import error in `httyp`. This will be changes when [#3](https://github.com/aelesia/npm-http/pull/3) is merged.

### DIFF
--- a/nodecg-io-reddit/extension/index.ts
+++ b/nodecg-io-reddit/extension/index.ts
@@ -1,0 +1,59 @@
+import { NodeCG } from "nodecg/types/server";
+import { ServiceClient } from "nodecg-io-core/extension/types";
+import { emptySuccess, success, Result } from "nodecg-io-core/extension/utils/result";
+import { ServiceBundle } from "nodecg-io-core/extension/serviceBundle";
+import RedditAPI from "reddit-ts";
+
+interface RedditServiceConfig {
+    clientId: string;
+    clientSecret: string;
+    username: string;
+    password: string;
+}
+
+export type RedditServiceClient = ServiceClient<RedditAPI>;
+
+module.exports = (nodecg: NodeCG) => {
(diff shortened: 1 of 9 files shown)


nodecg-io-reddit/extension/index.ts
```diff
@@ -0,0 +1,59 @@
...
+    async createClient(config: RedditServiceConfig): Promise<Result<RedditServiceClient>> {
+        const credentials = {
+            user_agent: "nodecg-io",
+            o2a: {
+                client_id: config.clientId,
+                client_secret: config.clientSecret,
+                username: config.username,
+                password: config.password,
+            },
+        };
```
daniel0611, line 45: It would be nice of converting our config to a reddit-ts config could be extracted into a seperate method since it also used in `validateConfig` and therefore could be reused.

daniel0611 (changes requested):
Only a couple nits, looking good so far 👍 . Thanks in advance!

nodecg-io-reddit/package.json
```diff
@@ -0,0 +1,35 @@
...
+    ],
+    "nodecg": {
+        "compatibleRange": "^1.1.1",
+        "bundleDependencies": {
+            "nodecg-io-core": "0.1.0"
+        }
+    },
+    "license": "MIT",
+    "devDependencies": {
+        "@types/node": "^14.6.4blueduck",
```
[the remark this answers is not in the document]
noeppi-noeppi: Whoops. Seems like I selected that somewhere else and accidentally pressed the middle mouse button so it was pasted there.

nodecg-io-reddit/extension/index.ts
noeppi-noeppi: The `Credentials` type is not exported.

daniel0611 (approved):
Thanks!

PR #138, opened 2020-11-14
Android service 

noeppi-noeppi:
Provides a way to send Intents to an android device that is connected via [adb](https://developer.android.com/studio/command-line/adb). Those will then be handles by the [nodecg-io-android](https://github.com/noeppi-noeppi/nodecg-io-android) app. Data will get sent back via the reverse port forwarding feature of adb. The goal is to provide an API similar to nodecg-io-intellij for android devices. The app (for now) requires Android 8. I might set it to Android 7 in the future.

This is not completed yet.


nodecg-io-android/extension/android.ts
```diff
@@ -0,0 +1,2586 @@
...
+        });
+        if (!result.available) {
+            throw new Error("The device does not implement wifi features.");
+        }
+        return new WifiManager(this);
+    }
+
+    equals(other: Android): boolean {
+        return this == other;
+    }
```
daniel0611, line 264: What's the intented use for this and the other `equals` methods since they are currently not used inside the service?

nodecg-io-android/package.json
```diff
@@ -0,0 +1,42 @@
...
+        }
+    },
+    "license": "MIT",
+    "devDependencies": {
+        "@types/node": "^14.14.13",
+        "nodecg": "^1.7.4",
+        "typescript": "^4.1.3"
+    },
+    "dependencies": {
+        "nodecg-io-core": "0.1.0",
```
daniel0611, line 38: Since you've started with the android service there were some changes to the template package.json and the nodecg-io label and the semver major operators are missing here. So add them to make this consistent across all services.
```suggestion
        "nodecg-io",
        "nodecg-bundle"
    ],
    "nodecg": {
        "compatibleRange": "^1.1.1",
        "bundleDependencies": {
            "nodecg-io-core": "^0.1.0"
        }
    },
    "license": "MIT",
    "devDependencies": {
        "@types/node": "^14.14.13",
        "nodecg": "^1.7.4",
        "typescript": "^4.1.3"
    },
    "dependencies": {
        "nodecg-io-core": "^0.1.0",
```

nodecg-io-android/extension/android.ts
noeppi-noeppi: When you for example get two contacts from two different sources (one from Recipient and one from the contact manager) they are not the same objects so I added those methods to compare them.

daniel0611 (approved):
Thanks!

PR #146, opened 2020-12-16
Added PubSub package and sample

### LINKED ISSUE #116: Add Twitch PubSub service
### Description
We should add a Twitch PubSub service to be able to react to e.g. channel points. Full list of supported events: https://dev.twitch.tv/docs/pubsub#topics. As a package we should use [twitch-pubsub-client](https://www.npmjs.com/package/twitch-pubsub-client).
I think the currently existing nodecg-io-twitch service should be renamed with this change to nodecg-io-twitch-chat inorder to make it clear that it only handles the chat and not other parts of Twitch.

derNiklaas:
closes #116 

uses: https://www.npmjs.com/package/twitch-pubsub-client

Note: I only tested the onRedemption function. I don't know if the others work, but they should work.


samples/pubsubsample/extension/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+module.exports = function(nodecg: NodeCG) {
+    nodecg.log.info("Sample bundle for discord started");
+
+    const pubsub = requireService<PubSubServiceClient>(nodecg, "pubsub");
+
+    pubsub?.onAvailable((client) => {
+        nodecg.log.info("PubSub client has been updated, adding handlers for messages.");
+        client.onSubscription(async (message) => {
+            const user = await message.getUser();
```
daniel0611, line 14: What would be the not deprecated way of doing this? Using a HelixApiClient and getting the user using the userId each time?

nodecg-io-pubsub/extension/index.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    async createClient(config: PubSubServiceConfig): Promise<Result<PubSubServiceClient>> {
+        const client = await PubSubServiceClient.createClient(config);
+
+        return success(client);
+    }
+
+    stopClient(_: PubSubServiceClient): void {
+    }
```
daniel0611, line 30: Sadly not (yet) possible :disappointed:, so add a comment to get rid of the linter warning. (Will also get funny atleast for me regarding #122)
```suggestion
    stopClient(_: PubSubServiceClient): void {
        // Not possible
    }
```

nodecg-io-pubsub/extension/index.ts
```diff
@@ -0,0 +1,31 @@
+import { NodeCG } from "nodecg/types/server";
+import { emptySuccess, Result, success } from "nodecg-io-core/extension/utils/result";
+import { ServiceBundle } from "nodecg-io-core/extension/serviceBundle";
+import { PubSubServiceClient } from "./pubSubClient";
+
+export { PubSubServiceClient } from "./pubSubClient";
```
daniel0611, line 6: nit: duplicate import

daniel0611 (changes requested):
Code looks pretty good so far :+1:.
However I would like the package to be renamed to `nodecg-io-twitch-pubsub` (and the service to `twitch-pubsub`) to make it clear that is for twitch as there may be other pubsub services that we may want to add in the future.
(chat will also be renamed to `nodecg-io-twitch-chat` for clarity but that can be done in another pull request and later).

Failing CI because of missing `express-transform-bare-module-specifiers@1.0.3` is fine. The older versions were removed from the npm registry and `1.0.4` is available.  The `package-lock.json` still uses `1.0.3` which results in the error. Will be fixed soon anyway as I want to bump all dependencies soon (maybe tomorrow if I have some time?).

nodecg-io-pubsub/extension/index.ts
derNiklaas: This is NOT a duplicate import. One says import, one says export (same as twitch-chat)

samples/pubsubsample/extension/index.ts
derNiklaas: I thought I couldn't use the user fields, but I just haven't pressed rebuild. 😅 
https://github.com/codeoverflow-org/nodecg-io/pull/146/commits/c725721f0527fb0e39bfaa306c003ec9fc7d811a should change this.

nodecg-io-pubsub/extension/index.ts
derNiklaas: Done

derNiklaas:
The last commit should've satisfy all requested changes.

nodecg-io-pubsub/extension/index.ts
daniel0611: Ooops didn't saw that, sorry

daniel0611 (approved):

daniel0611:
Thanks!

PR #153, opened 2020-12-29
Clear timeouts/intervals in sample bundles

daniel0611:
If the streamdeck was once available a timeout gets registered but never cleared.
This makes problems e.g. when the instance gets deleted: https://discord.com/channels/577412066994946060/577495348898168832/793590990475755550


daniel0611:
Now fixed for serial and websocket-client as well.

PR #163, opened 2021-01-08
Simplify importing of core

daniel0611:
Adds a `index.ts` to core that re-exports everything important. That way services and bundles can import everything they need by just doing this:
```typescript
import { ... } from "nodecg-io-core";
```

Previously bundle/service developers would need to know in which file is which class and provide the full path.
Everything obviously still can be imported by using the full path but if anything is moved this doesn't need fixing of the paths and therefore is prefered.

This is a follow up on #131 which simplified importing of services.

Docs update PR: https://github.com/codeoverflow-org/nodecg-io-docs/pull/42


derNiklaas:
IMO this should also be changed for every package/subproject.

daniel0611:
You mean in packages that are using nodecg-io e.g. skates-bundles or your private bundles? Yeah would propably be better.

PR #170, opened 2021-01-23
unified logging

MelanX:
- start of log messages are uppercase.
- `nodecg.log.info` ends with simple point `.`
- `nodecg.log.error` ends with exclamation mark `!`
- Sample bundles:
    - onAvailable says that it's updated and what it's doing now
    - onUnavailable says that it's unset
    - use service names in log


daniel0611 (approved):
:+1: Thanks

PR #197, opened 2021-03-01
Add a script to create a new service

noeppi-noeppi:


noeppi-noeppi:
The script does not work when using hyphens in a service name, so i converted it to a draft until fixed.

daniel0611 (commented):
Thanks!
A couple very small things but the rest is fine.

Can you also update the docs to use the script in the create a service guide?

.scripts/create-service.py
```diff
@@ -0,0 +1,177 @@
...
+
+    os.mkdir(f'samples/{sample_name}/extension')
+    with open(f'samples/{sample_name}/extension/index.ts', mode='w') as file:
+        file.writelines([
+            'import { NodeCG } from "nodecg/types/server";\n',
+            f'import {{ {service_name_c}Client }} from "nodecg-io-{service_name}";\n',
+            'import { requireService } from "nodecg-io-core";\n',
+            '\n',
+            'module.exports = function (nodecg: NodeCG) {\n',
+            f'    nodecg.log.info("{service_name_c} bundle started.");\n',
```
daniel0611, line 149: Actually most sample use a log here in this format: `Sample bundle for <name> started`. I think we should atlest make sure that it says that it is a sample bundle. Maybe add "sample" after the name? Can you also change this in the template while you're at it?

.scripts/create-service.py
```diff
@@ -146,31 +129,33 @@
...
-        file.writelines([
-            '<!-- Marker for build.py that there\'s no sample bundle. Remove this if you created one -->\n',
-            '\n',
-            f'# Documentation for {service_name_c}-service missing.\n',
-            '\n',
-            'You can help us [create it](../contribute/sample_documentation.md).'
-
-        ])
+    if not os.path.isdir('docs'):
+        print('Could not create documentation files. You should clone nocg-io-docs in the directory "docs".')
```
daniel0611, line 149:
```suggestion
        print('Could not create documentation files. You should clone nodecg-io-docs in the directory "docs".')
```

daniel0611 (approved):

PR #199, opened 2021-03-01
Twitch Addons service

noeppi-noeppi:
A service for the APIs of Better Twitch TV and FrankerFaceZ. The FFZ api is not completely covered but I think you can access almost everything. I'm not sure because I only have little knowledge about FFZ.

And before someone asks: I used the twitch channel of @derNiklaas in the sample because neither @sebinside nor @daniel0611 have any BetterTTV or FrankerFaceZ emotes added to their channels.


noeppi-noeppi:
Why does it fail?

samples/twitch-addons/extension/index.ts
```diff
@@ -0,0 +1,23 @@
+import { NodeCG } from "nodecg/types/server";
+import { TwitchAddonsClient } from "nodecg-io-twitch-addons";
+import { requireService } from "nodecg-io-core";
+
+module.exports = function (nodecg: NodeCG) {
+    nodecg.log.info("Twitch-addons bundle started.");
+
+    const twitch_addons = requireService<TwitchAddonsClient>(nodecg, "twitch-addons");
```
daniel0611, line 8: In TypeScript variables ussually (and always for this project) should be camel case and not with an underscore, if the name is freely chooseable and not already given e.g. because thats the name in a json response.. So this should be `twitchAddons` instead.

nodecg-io-twitch-addons/extension/index.ts
```diff
@@ -0,0 +1,79 @@
...
+        return success(client);
+    }
+
+    stopClient(_: TwitchAddonsClient): void {
+        this.nodecg.log.info("Successfully stopped twitch-addons client.");
+    }
+
+    removeHandlers(_: TwitchAddonsClient): void {
+        // TODO: Implement (optional)
+    }
```
daniel0611, line 54: This function can be fully removed because it is optional and not needed here

derNiklaas (changes requested):

daniel0611 (commented):
Thanks! Two very small nits but nothing of relevance.

nodecg-io-twitch-addons/extension/twitchAddonsClient.ts
```diff
@@ -0,0 +1,444 @@
...
+     */
+    is_group: boolean;
+    /**
+     * The display name (name with capitalisation) of the user
+     */
+    display_name: string;
+    /**
+     * The custom moderator badge url.
+     */
+    moderatorBade: string | null;
```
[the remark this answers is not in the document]
noeppi-noeppi: Not it's not because I just cast the data from the API to that object so both `moderatorBade` and `moderatorBadgeURL` would always be undefined...

PR #228, opened 2021-04-05
Fix `0.1.0` publishing problems

daniel0611:
The `0.1.0` release had some problems:

1. Missing `tslib` dependency on `nodecg-io-core`

	Since the beginning of nodecg-io we always had `importHelpers` set to true in `tsconfig.common.json`. Meaning TS is allowed to generate code that uses `tslib`. One day I removed `tslib` as a dependency because it was working without it, we had no code that needed `tslib` at that time.
	With #163 this changed because we had a star export, which needs `tslib`, but it worked fine in our dev environment because we have hoisted dependencies and various other packages depend on `tslib`.
	Because of this services cannot registered when they are installed in a non-hoisted environment and also bundles cannot use the core, unless the import everything using the full paths.

	Fixed in f89e593ff636467a67fa8e9ef04960e7b3315d2a by adding `tslib` as a dependency to the core.

2. Nanoleaf and Tiane exposed fetch/ws types but typings were in `devDependencies`

	In a perfect world (see issue 3) typings that are only used internally can be left in `devDependencies` and only typings that are exposed need to be in `dependencies`. This is also true for most of our published packages but not for Nanoleaf and Tiane.

	Nanoleaf exposed a fetch type in the following method: `NanoleafClient.getAllPanelInfo(): Promise<fetch.Response>` (maybe that function also should get a proper return type). Thus `@types/fetch` need to be a production dependency, which has been fixed in 8d8b318194c9b9ca2f98b6e22195883ed8b1230a.

	Tiane exposed the websocket client type in its public constructor of the Tiane service class: `constructor(websocket: WebSocket)`. This has been fixed in 63c9da50dc2fd8ee9cad94cb197a64c997498fef by making the constructor private and the `connectTiane` function a public static method of `Tiane`.

	Both of these issues weren't found in our dev environment because we always install `devDependencies`.

	The services are usable with this error but you cannot compile TypeScripts bundle depending on them unless you install the typings for these packages into your bundle.

3. Release of raw `.ts` files instead of generated `.d.ts` typings

	TypeScript can gather type information using the source `.ts` files or `.d.ts` which only contain the typings. While developing we always used the first way and it worked. The problem is that those files contain all imports, including those that are only used internally like `@types/crypto-js` (used by core) even when they are not exported. But in a production install we don't have those dependencies and we therefore cannot use the `.ts` files when e.g. building a bundle.

	Because of this bundles that are using ahk, curseforge, intellij, nanoleaf, tiane and xdotool cannot be compiled unless you temporarily install the `devDependencies`.

	The solution to this is to generate the `.d.ts` definitions and to not publish the source `.ts` files. This has been done in 665e311c967488212fd4a9177b6a1f21d2f51f1d and 3464b6d90073bef8eb27b0a1f3eab0a27922d408.

These changes will be cherry-picked to `release/0.1` and we will have patch release (`0.1.1`) published for the following packages:

- core (issue 1 and 3)
- ahk (issue 3)
- curseforge (issue 3)
- intellij (issue 3)
- nanoleaf (issue 2 and 3)
- tiane (issue 2 and 3)
- xdotool (issue 3)

The other services work as they are and because they use `^0.1.0` as a version selector for `nodecg-io-core` they will be able to work with the new version of the core.


daniel0611:
`0.1.1` for the affected packages has been released.

Issue #231, opened 2021-04-20
New bundles aren't in the bundle select even though they are loaded and mounted.


SteffoSpieler:
### Description
When creating new bundles the bundles are getting loaded by nodecg-io (and nodecg) as it mounts the bundle. (You can see it in the console.)
But the just created bundle just doesn't get displayed in the Bundle-Select on the dashboard.

### How to reproduce
1. Duplicate one random Bundle and rename it or just create a new one.
2. Do `npm i` and `npm run bsb`
3. Start the nodecg server.
4. Login to the nodecg-io dashboard and search for the just created bundle.

If you need more information just ask.
Also I already reinstalled nodecg-io 10 times. (multiple versions)

PR #232, opened 2021-04-20
Save config after all service instances have been loaded at startup

daniel0611:
Could be a solution for #231, was definitly not intented the way it was.

If you added a bundle you would first need to do an action that saved the config because otherwise it would be still the one from the last start without the new bundle. This was not noticied because usually you first create a new instance for the bundle and then you don't notice this behaviour.


daniel0611:
Backported to `release/0.1` and published as [`nodecg-io-core 0.1.2`](https://www.npmjs.com/package/nodecg-io-core/v/0.1.2)

PR #237, opened 2021-06-13
SQL service

### LINKED ISSUE #223: Add sql service
### Description
Add service to connect to mysql databases. There are multiple commonly used npm packages available.

### References
https://www.npmjs.com/package/knex

sebinside:
Close #223 by using knex with mysql as default client. Can be used with any other supported client, though.

Progress:
- [x] Create service client
- [x] Test service client
- [x] Add example bundle
- [x] Add some knex example code w/ explanations
- [x] Remove mysql from dependecy list, update comments


sebinside:
Still contains a reference to *mysql* in the `package.json` to make the service easy to get started with. Could be removed together with a change to the comment in the description in `nodecg-io-sql/package.json:4`.

daniel0611 (changes requested):
Please remove `nodecg-io-sql/package-lock.json`. It is created when using e.g. `npm install sqlite3` inside the sql service. Add your dependencies manually to the corresponding `package.json` file and the run `npm run bs` in the repository root (this also needs to be done to update the root `package-lock.jn`)

If you will add a explanation to the docs on how to install support for other sql dialects please use `npm install sqlite3 --no-save`. This will create no lockfile because for when a user manually installs a dependency these are not really required anyway.

I don't really care whether we want to include mysql by default. On the one hand it makes it easier to get started, on the other we could make the installation smaller for people not using mysql. I'm fine with both, do want you think is better. We still can change it after merging this PR.

daniel0611 (approved):

PR #245, opened 2021-08-18
Few changes

noeppi-noeppi:
* Added some methods to query file information for the android service using shell commands.
      This uses shell commands instead of the app as with shell commands it's possible  to access more files on the system that an app could on a non-rooted phone.
  * Clarify difference between `categoryID` and `sectionId` in the curseforge service.

### DIFF
--- a/nodecg-io-android/extension/android.ts
+++ b/nodecg-io-android/extension/android.ts
@@ -2629,7 +2629,7 @@ export class FileManager {
     }
 
     /**
-     * Downloads a file from the device. On some platforms, this gerts incredibly slow when used on
+     * Downloads a file from the device. On some platforms, this gets incredibly slow when used on
      * files larger than 6MB.
      */
     async download(device: string, local: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


nodecg-io-android/extension/android.ts
```diff
@@ -2582,6 +2590,138 @@ export type UsageStats = {
...
+
+    /**
+     * Gets some information about a file.
+     */
+    async file(path: string): Promise<string> {
+        return await this.android.rawAdb(["shell", "-b", path]);
+    }
+
+    /**
+     * Downloads a file from the device. On some platforms, this gerts incredibly slow when used on
```
daniel0611, line 2632:
```suggestion
     * Downloads a file from the device. On some platforms, this gets incredibly slow when used on
```

nodecg-io-android/extension/android.ts
```diff
@@ -2582,6 +2590,138 @@ export type UsageStats = {
...
+     * files larger than 6MB.
+     */
+    async download(device: string, local: string): Promise<void> {
+        await this.android.rawAdb(["shell", "pull", quote(device), quote(local)]);
+    }
+
+    /**
+     * Uploads a file to the device. On some platforms, this gerts incredibly slow when used on
```
daniel0611, line 2640:
```suggestion
     * Uploads a file to the device. On some platforms, this gets incredibly slow when used on
```

daniel0611 (approved):

PR #247, opened 2021-08-26
elgato-light service for Keylights and light stripes.

### LINKED ISSUE #118: Elgato Keylight integration.
@sebinside wanted a elgato keylight integation for nodecg-io. 

I found [this npm package](https://www.npmjs.com/package/elgato-light-api).

I could implement it, but I can't test it ('cause I don't have a keylight)

sebinside:
Closes #118. Does not use any external elgato light npm library because the're are all not up to date and we only need two HTTP requests...

### DIFF
--- a/nodecg-io-elgato-light/extension/elgatoLight.ts
+++ b/nodecg-io-elgato-light/extension/elgatoLight.ts
@@ -1,0 +1,91 @@
+import fetch from "node-fetch";
+import { LightData } from "./lightData";
+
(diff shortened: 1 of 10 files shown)


daniel0611 (approved):
I've refactored the `createLightData` method to take an object in 0a84989de14a4ec5c6803f04f122fa1881dd071b, hope you're fine with that. This makes it more clear what a call to it does because the parameter name is right in the call and you don't have to lookup its signature.
Otherwise this LGTM.

PR #249, opened 2021-08-28
Add config presets to core and dashboard. Add presets for midi-input

### LINKED ISSUE #109: Make device list of midi-input/-output available.
I think enabeling the midi-input and midi-output services to deliver a list of available devices would be beneficial since it would enable at least an easier configuration.

daniel0611:
Closes #109 

Currently only `midi-input` has presets. Other services still need to be done before this PR is merged.

How it works:

https://user-images.githubusercontent.com/30466471/131215363-a680621f-1835-46e8-888f-fdfcf40f0047.mp4

The preset drop-down is only visible if the service provides presets and is hidden for services that don't provide any otherwise the UI would be too cluttered.

### DIFF
--- a/nodecg-io-core/dashboard/crypto.ts
+++ b/nodecg-io-core/dashboard/crypto.ts
@@ -39,11 +39,11 @@ class Config extends EventEmitter {
 export const config = new Config();
 
 // Update the decrypted copy of the data once the encrypted version changes (if pw available).
-// This ensures that the decrypted data is kept uptodate.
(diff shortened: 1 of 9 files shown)


sebinside:
What a great feature!

daniel0611:
I've added presets to the Hue service with 2cac59846b53e05715f3034bbd3a612a2b654136 and removed the old discover feature. Would be nice if someone could test whether it works because I don't own a Hue bridge.

PR #268, opened 2021-10-18
Fix/npm workspaces

### LINKED ISSUE #263: Ditch lerna & add auto-updating for npm with dependabot
### Description
[Lerna](https://github.com/lerna/lerna) is currently used for managing the multiple packages.  

Especially [Hoisting](https://github.com/lerna/lerna/blob/main/doc/hoist.md) helps us to store common dependencies only once in the  `node_modules` folder. Without that feature the full installation of nodecg-io could become `>5GB` in file size and have 70 or so instances of TypeScript.  
Unfortunately hoisting is currently not supported by dependabot wich would break the `package-lock.json` on every update.

Also lerna is no longer maintained at the moment so it would be beneficial to ditch it in favour of npm workspaces.

Npm workspaces would have the benefit of working well together with dependabot (which should be used for automatic updates of dependencies) but we would have to find a way to execute the `build` script of all packets in parallel and in the right order.

### References
Dependabot for github actions: https://github.com/codeoverflow-org/nodecg-io/pull/262

TimTechDev:
Should fix #263

## Contents

- Remove `lerna`
- Add `npm-workspaces`
- New build process using typescript project references
- Remove webpack `ts-loder`
- Add update-paths.mjs
- Fix some TS errors
- Adapt pipeline

## Time savings

> On: Ubuntu 21.04 x86_64
> Using: AMD Ryzen 7 4700U (8) @ 2.000GHz

|        | Install | Bootstrap | Rebuild  |
| ------ | ------- | --------- | -------- |
| Before | 7.185s  | 14.386s   | 1m0.409s |
| After  | 18,262s | - - -     | 24,252s  |

### Before

```shell
$ time npm ci
real    0m7.185s
user    0m6.718s
sys     0m2.290s

$ time npm run bs
real    0m14.386s
user    0m18.274s
sys     0m3.085s

$ time npm run rebuild
real    1m0.409s
user    6m39.122s
sys     0m26.532s
```

### After

```shell
$ time npm ci
real    0m18,262s
user    0m24,543s
sys     0m5,145s

$ time npm run rebuild
real    0m24,252s
user    0m38,086s
sys     0m2,221s
```

### DIFF
--- a/services/nodecg-io-ahk/package.json
+++ b/services/nodecg-io-ahk/package.json
@@ -11,7 +11,7 @@
     "repository": {
         "type": "git",
         "url": "https://github.com/codeoverflow-org/nodecg-io.git",
-        "directory": "nodecg-io-ahk"
+        "directory": "services/nodecg-io-ahk"
     },
     "files": [
         "**/*.js",
--- a/services/nodecg-io-android/package.json
+++ b/services/nodecg-io-android/package.json
@@ -11,7 +11,7 @@
     "repository": {
         "type": "git",
         "url": "https://github.com/codeoverflow-org/nodecg-io.git",
-        "directory": "nodecg-io-android"
+        "directory": "services/nodecg-io-android"
     },
     "files": [
         "**/*.js",
--- a/services/nodecg-io-artnet/package.json
+++ b/services/nodecg-io-artnet/package.json
@@ -10,7 +10,7 @@
     "repository": {
         "type": "git",
         "url": "https://github.com/codeoverflow-org/nodecg-io.git",
-        "directory": "nodecg-io-artnet"
+        "directory": "services/nodecg-io-artnet"
     },
     "files": [
         "**/*.js",
--- a/services/nodecg-io-curseforge/package.json
+++ b/services/nodecg-io-curseforge/package.json
@@ -10,7 +10,7 @@
     "repository": {
         "type": "git",
         "url": "https://github.com/codeoverflow-org/nodecg-io.git",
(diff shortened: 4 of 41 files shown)


Tim-Tech-Dev:
> monaco-editor in the core and debug dashboard cannot be loaded because they are not inside the node_modules directory of the corresponding dashboard anymore.

Well, I can build and use the dashboard in the current branch, so I can't find this problem. May you explain the issue more clearly? Thank you.

daniel0611:
> > monaco-editor in the core and debug dashboard cannot be loaded because they are not inside the node_modules directory of the corresponding dashboard anymore.
> 
> Well, I can build and use the dashboard in the current branch, so I can't find this problem. May you explain the issue more clearly? Thank you.

Hmmm, for me it fails because `http://localhost:9090/bundles/nodecg-io-debug/monaco/monaco-editor/min/vs/loader.js
` returns 404. Maybe it is still in the `node_modules` directory for you. Can you try to remove all `node_modules` directories and re-running `npm install` and check if then pops up? Otherwise I can fix it and do a PR aswell.

> Why does the lint fail?

Previously running `npm install` or `npm ci` would only install the dependencies that are defined in the root `package.json` and all other dependencies would be installed when `npm run bootstrap` was run.
With npm workspaces `npm install` already installs all dependencies, including those in all services. Some services like midi use native modules which require some os packages which are not installed in the lint workflow. To fix this you can copy the `apt install ...` step from the CI to the lint workflow.

PR #287, opened 2021-11-04
Feature: Update nodecg-io-streamelements to include test events

derNiklaas:
This PR adds a new field ``handleTestEvents`` to the SE config which controls whether it should also fire test events.

Also bumps the version to ``0.2.1`` for the sample and nodecg-io-streamelements

### DIFF
--- a/samples/streamelements-events/package.json
+++ b/samples/streamelements-events/package.json
@@ -1,6 +1,6 @@
 {
     "name": "streamelements-events",
-    "version": "0.2.0",
+    "version": "0.2.1",
     "private": true,
     "nodecg": {
         "compatibleRange": "^1.1.1",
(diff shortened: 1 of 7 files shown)


derNiklaas:
Also yes, we require a 2.x.x version of the socket-io-client because streamelements hasn't updated their service and v3+ doesn't work with 2.x.x connections

samples/streamelements-events/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "streamelements-events",
-    "version": "0.2.0",
+    "version": "0.2.1",
```
daniel0611, line 3: I don't really mind having `0.2.1` here, but `0.2.0` wasn't released yet and it would be kinda weird if we start with `0.2.1` when we're going to release `0.2` sometimes in the future. I think I'm gonna push some features of the 0.2 milestone into 0.3 so we can release all the new services.

daniel0611:
Dang :facepalm:. How many times did I accidentially upgrade the socket.io lib to 3.0.0 or higher? Propably too often, lol.

daniel0611 (commented):
Thanks in advance for this useful PR!

daniel0611 (approved):
Thanks!

PR #324, opened 2021-11-20
Add StreamElementsTestEventData interface for correct "onTest" types.

SteffoSpieler:
StreamElements sends out `name` instead of `username` in the onTest Event.
This adds the correct type for that specific event.


daniel0611 (approved):
Thanks. Do you want a bug fix release for this (`nodecg-io-streamelements 0.2.1`) or is waiting till `0.3.0` fine?

Issue #325, opened 2021-11-21
Failed to mount nodecg-io-irc extension: Cannot find module 'irc-colors'


SteffoSpieler:
### Description

When starting nodecg-io/nodecg I get this error.

### How to reproduce

Can't give exact repoduction steps, as I just keep `git pull`ing the repository.

1. Start nodecg with `nodecg start`
2. See error.

### Error Log 
 
```
warn: [nodecg/lib/server/extensions] Failed to mount nodecg-io-irc extension:
Error: Cannot find module 'irc-colors'
Require stack:
- F:\NodeCG\nodecg\nodecg-io\services\nodecg-io-irc\node_modules\irc\lib\parse_message.js
- F:\NodeCG\nodecg\nodecg-io\services\nodecg-io-irc\node_modules\irc\lib\irc.js
- F:\NodeCG\nodecg\nodecg-io\services\nodecg-io-irc\extension\index.js
- F:\NodeCG\nodecg\lib\server\extensions.js
- F:\NodeCG\nodecg\lib\server\index.js
- F:\NodeCG\nodecg\index.js
- C:\Users\User\AppData\Roaming\npm\node_modules\nodecg-cli\dist\commands\start.js
- C:\Users\User\AppData\Roaming\npm\node_modules\nodecg-cli\dist\commands\index.js
- C:\Users\User\AppData\Roaming\npm\node_modules\nodecg-cli\dist\index.js
- C:\Users\User\AppData\Roaming\npm\node_modules\nodecg-cli\dist\bin\nodecg.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (F:\NodeCG\nodecg\nodecg-io\services\nodecg-io-irc\node_modules\irc\lib\parse_message.js:1:17)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
[...]

PR #327, opened 2021-11-22
feat/NodeCG-CI

Tim-Tech-Dev:
Minimize runtime errors by actually trying to run our
framework with NodeCG. This CI will run on push and
pull requests to the main branch.

As suggested in the discussion following #325.


Tim-Tech-Dev:
The action is failing because NodeCG wants to write to folders it should not.

https://github.com/Tim-Tech-Dev/nodecg-io/runs/4292590810

daniel0611:
NodeCG only seems to work when given a absolute path to the bundle directories, at least if the bundle directories are in nested sub directories. To fix this we can provide the full path using the main directory you currently already log at the beginning: https://github.com/daniel0611/nodecg-io/commit/b37785373df4f90d708e3eca93bc4836dbb985ac.
I would recommend integrating this check into our current CI check. Both check that our project can be compiled and having two seperate workflows for this don't really provide an advantage other than the workflow without the NodeCG run check runs a bit faster.

Tim-Tech-Dev:
I would suggest using the `$GITHUB_WORKSPACE` [environment variable](https://docs.github.com/en/actions/learn-github-actions/environment-variables). So we don't rely on the path of the runner.

Tim-Tech-Dev:
I missed the `working-directory` values in my commit.

Tim-Tech-Dev:
Added the NodeCG runt to the publishing requirements.

daniel0611 (approved):
Thank you very much

PR #384, opened 2021-12-19
Downgrade monaco-editor from 0.31.1 to 0.30.1

daniel0611:
Reverts d1da66337a6c0c141d174bd5aaa8a4be4baf545a / #377

There is some change with `0.30.0` that results in a build output by webpack that
is incompatible with the Chromium version that is used by custom OBS Docks.

I don't know what the exact problem is yet with this specific update,
this PR is just here to fix this bug till I figure out if we can make
the new version compatible with OBS Docks or not.

### DIFF
--- a/nodecg-io-core/dashboard/package.json
+++ b/nodecg-io-core/dashboard/package.json
@@ -24,7 +24,7 @@
         "style-loader": "^3.3.1",
         "css-loader": "^6.5.1",
         "monaco-editor-webpack-plugin": "^7.0.1",
-        "monaco-editor": "^0.31.1"
+        "monaco-editor": "^0.30.1"
     },
     "dependencies": {
         "crypto-js": "^4.1.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -9244,7 +9244,8 @@
             "version": "0.31.1",
             "resolved": "https://registry.npmjs.org/monaco-editor/-/monaco-editor-0.31.1.tgz",
             "integrity": "sha512-FYPwxGZAeP6mRRyrr5XTGHD9gRXVjy7GUzF4IPChnyt3fS5WrNxIkS8DNujWf6EQy0Zlzpxw8oTVE+mWI2/D1Q==",
-            "dev": true
+            "dev": true,
+            "peer": true
         },
         "node_modules/monaco-editor-webpack-plugin": {
             "version": "7.0.1",
@@ -14030,7 +14031,7 @@
             "devDependencies": {
                 "clean-webpack-plugin": "^4.0.0",
                 "css-loader": "^6.5.1",
-                "monaco-editor": "^0.31.1",
+                "monaco-editor": "^0.30.1",
                 "monaco-editor-webpack-plugin": "^7.0.1",
                 "style-loader": "^3.3.1",
                 "typescript": "^4.5.4",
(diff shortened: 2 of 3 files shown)


daniel0611:
Update on this: monaco `0.30.1` uses `es2018` or lower, monaco `0.31.0` switched to `esnext` (in https://github.com/microsoft/monaco-editor/commit/c419101f73d1c086bd5345448296bffd60892ca9) which is too new for the chromium that is embedded in OBS.

This means we'll have to transpile the JS to `es2018` in our build step. To do this we could use babel in combination with webpack. We could also move to something like [esbuild](https://esbuild.github.io/) which would make our builds faster.
I guess we'll stay on webpack for now because I don 't want to interfere with the work on the vue gui branch by @Tim-Tech-Dev.

Tim-Tech-Dev:
> I guess we'll stay on webpack for now because I don 't want to interfere with the work on the vue gui branch by @Tim-Tech-Dev.

It seems like a switch to `esbuild` would not really harm my progress. Most of what I have done so far isn't really webpack specific. We may move to `esbuild` should be a relatively easy fix.

daniel0611:
Okay, great. I'll have a look as esbuild and try to migrate to it. *Should* be pretty easy.

Tim-Tech-Dev:
I already tried it and it resulted in this:
Screenshot 2021-12-28 at 12-15-06 nodecg-io-core

daniel0611:
Have you added esbuild using their docs or have you used the [monaco esbuild sample](https://github.com/microsoft/monaco-editor/tree/main/samples/browser-esm-esbuild)? Monaco always needs some special options. I guess their sample should work as intented, but haven't tried it yet.

Tim-Tech-Dev:
I adopted the docs but, but I will take a look at the sample.

Tim-Tech-Dev:
The sample does not help either.

daniel0611:
Hmm, sad. Getting monaco working is always a pita. Can you push your current work to a branch on your fork? I guess I can take a look at it and try to figure it out. Otherwise I guess we'll just stay with webpack.

Tim-Tech-Dev:
https://github.com/Tim-Tech-Dev/nodecg-io/tree/feat/WIP-esbuild

Tim-Tech-Dev:
It seems to be that some functions are just not available in the DOM.

daniel0611:
As far as I've seen there currently two problems.
1. Monaco is failing to load its workers. To fix that you just need to prepend `dist/` in the paths inside the MonacoEnvironment you've copied from the sample because it assumes they are in the same directory than the html file but for us this is (currently) not the case. The html is in the dashboard directory and the workers are in the `dist` directory.

2. The buttons and handlers cannot be executed. This is propably why the save button won't work (neither does the login button which catched me).
Webpack had an option `libraryTarget` which controlls what happens to the exports in `main.ts`. It was set to `global` which means all exports in `main.ts` were assigned to the global object which is then used by the `onclick` (and other) handlers in the html file. Esbuild defaults to just run the code but not set the exports to the global object which is why the error says that e.g. `loadFramework()` is not defined.
I've searched a bit but it seems that esbuild does not have such option. This means we need to modify the `main.ts` to assign the functions to the global `window` object. Temporary workarround: replace `main.ts` with this
```ts
import { loadFramework } from "./authentication";
import { renderBundleDeps, renderInstanceSelector, setSelectedServiceDependency, unsetAllBundleDependencies } from "./bundles";
import { createInstance, deleteInstance, onInstanceSelectChange, saveInstanceConfig, selectInstanceConfigPreset } from "./serviceInstance";
global.loadFramework = loadFramework;
global.onInstanceSelectChange = onInstanceSelectChange;
global.createInstance = createInstance;
global.saveInstanceConfig = saveInstanceConfig;
global.deleteInstance = deleteInstance;
global.selectInstanceConfigPreset = selectInstanceConfigPreset;
global.renderBundleDeps = renderBundleDeps;
global.renderInstanceSelector = renderInstanceSelector;
global.setSelectedServiceDependency = setSelectedServiceDependency;
global.unsetAllBundleDependencies = unsetAllBundleDependencies;
```
This does not pass type checking and there is propably a better way to do it, but for now it works.

daniel0611:
We could also add all the handlers from TypeScript instead of defining them in the html, but I'm not sure if that's good coding style or not.

Tim-Tech-Dev:
https://github.com/Tim-Tech-Dev/nodecg-io/commit/51a74a48340ebf156eaa9b4e160e78d7dc4544ea

PR #417, opened 2022-01-08
Fix philipshue service after `is-ip` update

Tim-Tech-Dev:


daniel0611 (approved):
Only importing the function makes this definitely cleaner, thanks

PR #421, opened 2022-01-08
chore:Format configs & scripts

Tim-Tech-Dev:
- Add JSDoc types to `jest.config.js`, `.eslintrc.js` & `.prettierrc.js`
- Set more prettier configs explicitly to circumvent an issue I experienced with the prettier
  vscode extension, where it would format some stuff based on my global preferences
  and not the defaults.
- Use Set in `esbuild.config.js`
- Format all js and ts files
- Use module syntax in jest config
- Automatically add packages with jest config to root jest projects
- Add .mjs extension to files formatted by prettier
- Enable type checking JS files project wide

    Tim

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,13 +1,11 @@
+/** @type {import("eslint").ESLint.Options} */
 module.exports = {
     parser: "@typescript-eslint/parser",
     parserOptions: {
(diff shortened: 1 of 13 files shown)


daniel0611 (approved):
Thanks!

PR #422, opened 2022-01-09
Make OAuth redirect URLs match nodecg's baseUrl & protocol

Dan-Shields:
- Adds an optional config option `httpsRedirect` to spotify and googleapis services
- Makes googleapis service use NodeCG's `config.baseURL` for redirect URL as the spotify service already does

### DIFF
--- a/services/nodecg-io-googleapis/extension/index.ts
+++ b/services/nodecg-io-googleapis/extension/index.ts
@@ -11,6 +11,7 @@ interface GoogleApisServiceConfig {
     clientSecret: string;
     refreshToken?: string;
     scopes?: string | string[];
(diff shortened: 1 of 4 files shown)


daniel0611:
Could you please add the `httpsRedirect` option to the json schema that's located in each service directory? That way it will be automatically suggesting when using CTRL + Space in the dashboard with a description of the option.

Dan-Shields:
> Could you please add the `httpsRedirect` option to the json schema that's located in each service directory? That way it will be automatically suggesting when using CTRL + Space in the dashboard with a description of the option.

Done, good catch!

daniel0611 (approved):
Thanks!

PR #424, opened 2022-01-09
Derive encryption key in dashboard using Argon2

daniel0611:
See https://github.com/codeoverflow-org/nodecg-io/pull/408#issuecomment-1006896276.

After doing some research I found out that a key derivation function like PBKDF2 is probably better suited for this than a plain hashing algorithm.
CryptoJS actually already does this when a password is provided to the `AES.encrypt` function.
We now just do the key derivation ourselves in the dashboard, only transmit the key, and pass the key to CryptoJS.
Because we're now ourselves managing the encryption key we also need to manage the salt and initialization vector.

All old configurations will be automatically migrated on the first login.

Todo list:
- [x] First implementation
- [x] Code cleanup and documentation
- [x] Fix and update tests

Any feedback is highly appreciated.


nodecg-io-core/extension/persistenceManager.ts
```diff
@@ -55,18 +69,28 @@ export function decryptData(
     }
 }
 
-export function encryptData(data: PersistentData, password: string | crypto.lib.WordArray): [string, string] {
+/**
+ * Encrypts the passed data objedt using the passed encryption key.
```
Tim-Tech-Dev, line 73:
```suggestion
 * Encrypts the passed data object using the passed encryption key.
```

nodecg-io-core/extension/persistenceManager.ts
```diff
@@ -48,11 +69,71 @@ export function decryptData(cipherText: string, password: string): Result<Persis
...
+}
+
+/**
+ * Derives a key suitable for encrypting the config from the given password.
+ *
+ * @param password the password from which the encryption key will be derived.
+ * @param salt the salt that is used for key derivation.
+ * @returns a hex encoded string of the derived key.
+ */
+export function deriveEncryptionKey(password: string, salt: string): string {
```
sebinside, line 93: Is this really *deriving* or more precisely calculating?

nodecg-io-core/extension/persistenceManager.ts
```diff
@@ -48,11 +69,71 @@ export function decryptData(cipherText: string, password: string): Result<Persis
...
+    const saltWordArray = crypto.enc.Hex.parse(salt);
+
+    return crypto
+        .PBKDF2(password, saltWordArray, {
+            keySize: 256 / 32,
+            iterations: 5000,
```
sebinside, line 99: Is 5000 iterations like normal, or your approach, where does this originate, what does it mean? a comment might be a good fit for these settings

nodecg-io-core/extension/persistenceManager.ts
```diff
@@ -292,12 +381,36 @@ export class PersistenceManager {
             if (bundles.length > 0) {
                 try {
                     this.nodecg.log.info("Attempting to automatically login...");
-                    const loadResult = await this.load(password);
+
```
sebinside, line 384: This code duplication with `crypto.ts` is required because there is no way to serve the same functionality in the extension and the dashboard? Then I would propose to at least add a note to minimize breaking this in evolution (which could probably mess up configs...)

sebinside (changes requested):
Although I welcome the changes to increase security, I am not qualified enough to state whether this is correctly realized. Thank for testing, I hope migration will work 😁

nodecg-io-core/extension/persistenceManager.ts
daniel0611: Well... this function is just a wrapper around PBKDF2 (Password-Based *Key Derivation* Function 2). So I assume this is deriving?

nodecg-io-core/extension/persistenceManager.ts
daniel0611: Fixed in https://github.com/codeoverflow-org/nodecg-io/pull/424/commits/f3ee48772998a8ea54d89b38738ae34dfa4b6352

nodecg-io-core/extension/persistenceManager.ts
daniel0611: See https://github.com/codeoverflow-org/nodecg-io/pull/424/commits/b504cc5a9439c6f6e052976664dc64738f842562

daniel0611:
Because of the problems with this PBKDF2 implementation (see https://github.com/codeoverflow-org/nodecg-io/commit/b504cc5a9439c6f6e052976664dc64738f842562) I'm thinking about exploring argon2 for this. It has a [WebAssembly implementation](https://github.com/antelle/argon2-browser) available and thus has not the performance problems that running PBKDF2 in JS has. Additionally argon2 seems to be [more secure](https://cryptobook.nakov.com/mac-and-key-derivation/argon2#:~:text=%E2%80%8BArgon2%20is%20modern%20ASIC,for%20CPU%20and%20RAM%20usage). If we're changing the config layout we might as well do it future-proof. However this is currently only a idea and I will need to explore that possibility and decide whether or not we should do it.

I still should add at least one unit test for migration. It has always worked for my various configs but I don't want to have the migration suddenly break in future nodecg-io versions.

Additionally I should write the security assumptions that nodecg-io makes into the docs before merging this PR.

I hopefully should have some spare time in about two weeks to do all this, handle some maintenance tasks like removing the broken curseforge service and release 0.3.0.

daniel0611:
This PR is *finally* ready to be merged. @sebinside could you please re-review this?

PR #434, opened 2022-01-17
Fix enum suggestions in editor

cesmec:
Suggestions for enum types did not work correctly because `objectDeepCopy` turned Arrays into Objects.
This fixes for example the `lightType` property in elgato-light configs, as well as many properties of the serial service config.


daniel0611 (approved):
Thanks!

PR #465, opened 2022-02-11
Use https to clone the reddit dependencies

noeppi-noeppi:
While fixing https://github.com/noeppi-noeppi/npm-reddit-ts/pull/1, I found out that yarn apparently has problems cloning the git repo via ssh. I don't know whether nodecg-io even can be used with yarn but I think making it use https explicitly should not be a problem, so I don't see a reason not to.


daniel0611 (approved):
Thanks! NodeCG bundles can use released versions of nodecg-io through yarn, as far as I know.

PR #494, opened 2022-03-02
Add replicant with last events to streamelement service

daniel0611:
This is my take on replicants (#65) that I've shared on Discord before.
The bundle that wants a replicant creates it in the extension of the bundle and then calls a method on the service client that sets up handlers to feed data into it. That way the replicant is declared in the context of the bundle using the service, the name of the replicant is freely choosable and, most important, static.

To start I've added a replicant to the streamelements service that contains the last sub, last follow, last tip, etc. To show how the replicant can be used I've added a simple graphic to the `streamelements-events` sample bundle that gets the data from the replicant and displays it using vue.

TODO:
- [x] Initial implementation for streamelements and changes to sample bundle
- [ ] Fix current workaround by properly setting an alias to the esm version for vue in the graphic.
- [ ] Documentation about usage
- [ ] Add replicants to more services

### DIFF
--- a/samples/streamelements-events/graphics/index.ts
+++ b/samples/streamelements-events/graphics/index.ts
@@ -1,0 +1,32 @@
+/// <reference types="nodecg-types/types/browser" />
+import { createApp, defineComponent } from "vue/dist/vue.esm-bundler.js";
+import type { StreamElementsReplicant } from "nodecg-io-streamelements";
+
+const replicant = nodecg.Replicant<StreamElementsReplicant>("streamelements");
+
+const mainComponent = defineComponent<{}, {}, { se: StreamElementsReplicant }>({
+    data() {
+        return {
+            se: {},
+        };
+    },
+    created() {
+        replicant.on("change", (newVal) => {
+            this.se = newVal;
+        });
+    },
+    computed: {
+        subTier() {
+            const sub = this.se.lastSubscriber;
+            if (!sub || !sub.data.tier) return undefined;
+
+            if (sub.data.tier === "prime") {
+                return "Twitch Prime";
+            }
+
+            return `Tier ${Number.parseInt(sub.data.tier) / 1000}`;
+        },
+    },
+});
+
+createApp(mainComponent).mount("#app");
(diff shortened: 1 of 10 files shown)


samples/streamelements-events/graphics/index.ts
```diff
@@ -0,0 +1,32 @@
+/// <reference types="nodecg-types/types/browser" />
+import { createApp, defineComponent } from "vue";
+import type { StreamElementsReplicant } from "nodecg-io-streamelements";
+
+const replicant = nodecg.Replicant<StreamElementsReplicant>("streamelements");
+
+const mainComponent = defineComponent<unknown, unknown, { se: StreamElementsReplicant }>({
```
sebinside, line 7: Minor comment: I would always prefer non-abbreviated variable names, not only in an illustrative example

samples/streamelements-events/graphics/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    computed: {
+        subTier() {
+            const sub = this.se.lastSubscriber;
+            if (!sub || !sub.data.tier) return undefined;
+
+            if (sub.data.tier === "prime") {
+                return "Twitch Prime";
+            }
+
+            return `Tier ${Number.parseInt(sub.data.tier) / 1000}`;
```
sebinside, line 27: Magic number, what means 1000?

sebinside (approved):
I really like this approach regarding our discussion in #65 (I would even propose that this example could close that issue, too). The stream elements service is a good choice and the example is easy to understand and follow, regarding replicant usage and overall low-code possibilities with nodecg-io.

samples/streamelements-events/graphics/index.ts
daniel0611: StreamElements provides the tiers in multiples of 1000: https://github.com/codeoverflow-org/nodecg-io/blob/main/services/nodecg-io-streamelements/extension/StreamElementsEvent.ts#L72
I should probably add an explanation here and in the extension part of the sample where it is the same.
Thanks for your review!

PR #531, opened 2022-03-30
Priorize exact midi device name match in case of an ambiguous pattern

daniel0611:
Fix for https://discord.com/channels/577412066994946060/577495348898168832/958791971386843147

### DIFF
--- a/services/nodecg-io-midi-input/extension/index.ts
+++ b/services/nodecg-io-midi-input/extension/index.ts
@@ -27,11 +27,15 @@ class MidiService extends ServiceBundle<MidiInputServiceConfig, MidiInputService
                     devices.push(device);
                 }
             });
+
             if (devices.length === 0) {
-                return error("No device matched the configured pattern!");
+                return error("No device matched the configured pattern.");
             }
-            if (devices.length > 1) {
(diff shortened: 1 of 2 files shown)


voruti:
Tested in new installation. Works for me. Thanks!

services/nodecg-io-midi-input/extension/index.ts
```diff
@@ -48,11 +52,18 @@ class MidiService extends ServiceBundle<MidiInputServiceConfig, MidiInputService
...
-                    deviceName = device;
-                }
-            });
+            const allDevices = easymidi.getInputs();
+            if (allDevices.includes(config.device)) {
+                // If we have a device with the correct name we use that device.
+                deviceName = config.device;
+            } else {
+                // Otherwise we find a device which contains the pattern.
+                easymidi.getOutputs().forEach((device) => {
```
voruti, line 61: Typo: Should be
```js
easymidi.getInputs().forEach((device) => {
```

daniel0611: :facepalm:. Fixed with https://github.com/codeoverflow-org/nodecg-io/pull/550

PR #673, opened 2022-07-14
Fix undefined error message in obs service

ExtremTechniker:
Now outputting correct error when the connection to obs can't be established

Before:
```
error: [nodecg-io-core] The "obs" service with the name "obs" produced an error while creating a client: undefined
```
After:
```
error: [nodecg-io-core] The "obs" service with the name "obs" produced an error while creating a client: connect ECONNREFUSED 127.0.0.1:4444
```


daniel0611 (approved):
Thanks!

PR #708, opened 2022-07-31
fix: Use correct eventName and event function name in Stream Elements

SteffoSpieler:
I missed the usage of a wrong eventName and with that an inconsistant naming of the onTestSubscriber function.


daniel0611 (approved):
Thanks for the fix!

PR #761, opened 2022-09-04
Fix OpenTTS getWavData, format and add doc comments

daniel0611:
`getWavData` prepended the base URL again before the already valid URL, making it invalid.

Also adds docs comments to the OpenTTS service and runs prettier to format the code.


SteffoSpieler (approved):
LGTM

PR #894, opened 2023-03-12
Patch version of twitch-auth package in published tarballs

daniel0611:
Makes twitch services work in generated bundles of nodecg-io-cli development environments. See https://github.com/codeoverflow-org/nodecg-io/discussions/893#discussioncomment-5280262 for a more detailed explanation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,6 +166,20 @@ jobs:
             - name: Extract compilation results
               run: tar xvf compilation-results.tar.gz
 
+              # All twitch services depend on the utility package `nodecg-io-twitch-auth`.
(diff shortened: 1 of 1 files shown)


daniel0611:
@MTRNord thanks again for pointing out this issue.
The currently published tarballs include the working fix.

However your local npm very likely still has the old tarball cached, so if you want to use it you may need to clear your npm cache using `npm cache clean --force`

MTRNord:
> @MTRNord thanks again for pointing out this issue.
> The currently published tarballs include the working fix.
> 
> However your local npm very likely still has the old tarball cached, so if you want to use it you may need to clear your npm cache using `npm cache clean --force`

Thanks a lot for the fix 👍 :)

PR #991, opened 2023-08-21
feat(gametts): Add gametts service

SteffoSpieler:
Add a service for the german gametts docker container.

Also the opentts sample graphics has been fixed.


daniel0611 (approved):
Pair programmed. LGTM

Issue #1030, opened 2023-10-29
Fix MIDI services not working in GitHub Actions CI


daniel0611:
### Description
Currently, the CI of nodecg-io is broken because of the MIDI services which cannot start and break the NodeCG integration test as they cannot be loaded successfully.
This might be a permission problem in the CI VM or some missing package.
Sadly it is not known what GitHub changed since the tests previously run on GitHub Actions, run locally, and only have started failing at some time point.

### How to reproduce
Let CI run and view the NodeCG integration test results.

### Error Log 
 
```
2023-10-28 21:12:56 - info: [nodecg-io-midi-input] midi-input bundle started.
ALSA lib seq_hw.c:466:(snd_seq_hw_open) open /dev/snd/seq failed: No such file or directory

MidiInAlsa::initialize: error creating ALSA sequencer client object.
```

TimTechDev:
I would say that the issue is probably caused by the switch by GitHub to the action runner v4. Additionally, there is a new Ubuntu base image.

Here is a diff (with cleanup) over the logs between the change:
```diff
run-6133176760 run-6209625847 
1c1
< # Start time 2023-09-09T21:09:49.8807687Z
---
> # Start time 2023-09-22T20:51:55.7951375Z
6,7c6,7
< Job is about to start running on the hosted runner: GitHub Actions 4 (hosted)
< Current runner version: '2.308.0'
---
> Job is about to start running on the hosted runner: GitHub Actions 3 (hosted)
> Current runner version: '2.309.0'
15,17c15,17
< Version: 20230903.1.0
< Included Software: https://github.com/actions/runner-images/blob/ubuntu22/20230903.1/images/linux/Ubuntu2204-Readme.md
< Image Release: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20230903.1
---
> Version: 20230917.1.0
> Included Software: https://github.com/actions/runner-images/blob/ubuntu22/20230917.1/images/linux/Ubuntu2204-Readme.md
> Image Release: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20230917.1
1041d1040
< Get:6 https://packages.microsoft.com/ubuntu/22.04/prod jammy InRelease [3611 B]
1044,1057c1043,1057
< Hit:7 https://ppa.launchpadcontent.net/ubuntu-toolchain-r/test/ubuntu jammy InRelease
[...]

PR #1046, opened 2023-12-01
ci: Temp fix nodecg job by downgrading ubuntu

TimTechDev:
Move nodecg job to ubuntu 20.04, because 22.04 is currently missing some ALSA kernel modules. These are needed to use the midi services.

Resolves codeoverflow-org/nodecg-io#1030


daniel0611:
I've chosen #1049 over this PR as mentioned in https://github.com/codeoverflow-org/nodecg-io/issues/1030#issuecomment-1836068305

PR #1049, opened 2023-12-01
ci: Temp fix nodecg job testing midi on 20.04

TimTechDev:
22.04 is currently missing some ALSA kernel modules, which are needed to use the midi services.
So we use matrix execution to test all but the midi services on Ubuntu 22.04 and everything on 20.04.

The midi services are only tested on 20.04!

Resolves codeoverflow-org/nodecg-io#1030


daniel0611 (approved):
Thanks!

